PR #19, opened 2024-03-25
[Doop] Add mock data and remove useEndlessScroll hook

hodanoori:
plus:
1.  implementing mock data handling using our utils lib
2. debounce setSearchterm to avoid unnecessary re-renders


doop/ui/src/components/AppContent.jsx
```diff
@@ -16,10 +16,39 @@ import { fetchData } from "../lib/apiClient"
...
+        ...{ mock: true },
+      })
+        .then((response) => {
+          if (!response.ok) {
+            addMessage({
+              variant: "error",
+              text: parseError(e.message),
+            })
+          }
+          return response.json()
```
andypf, line 45: It means that an attempt will be made to convert the response to JSON in any case, even if response.ok is false?

doop/ui/src/components/filters/FilterSelect.jsx
```diff
@@ -87,9 +97,7 @@ const FilterSelect = () => {
       <SearchInput
         className="w-96"
         value={searchValue || ""}
-        onChange={(e) => {
-          setSearchTerm(e.target.value)
-        }}
+        onChange={(value) => handleSearchChange(value)}
```
andypf, line 100: 👍

doop/ui/src/lib/apiClient.js
```diff
@@ -5,13 +5,6 @@
 
 // Example fetch call. Adjust as needed for your API
 export const fetchData = ({ queryKey, meta }) => {
-  if (meta.mock === "true" || meta.mock === true) {
```
andypf, line 8: Perhaps the code was well placed here?

doop/ui/db.json
```diff
@@ -0,0 +1,452 @@
+{
```
andypf, line 1: 👍

doop/ui/src/components/AppContent.jsx
hodanoori: I tested this, return response.json() is needed in order to catch the result in the next then, as this promise is pending here and not resolved yet.

doop/ui/src/components/AppContent.jsx
```diff
@@ -16,10 +16,39 @@ import { fetchData } from "../lib/apiClient"
...
+        })
+        .then((result) => {
+          setData(result)
+        })
+    }
+  }, [isMock])
 
   // LOAD DATA
   const dataRequest = useQuery({
```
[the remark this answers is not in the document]
hodanoori: I handle this adding enabled:!isMock prop here.

andypf (approved):

PR #22, opened 2024-04-02
fix(dev-env):

uwe-mayer:
* Add boostrapped resources
    * Set correct image tag


dev-env/bootstrap/teams.yaml
```diff
@@ -0,0 +1,94 @@
...
+kind: TeamMembership
+metadata:
+  name: test-team-2
+  namespace: test-org
+spec:
+  members:
+  - id:        "01234"
+    firstName: "Jane"
+    lastName:  "Doe"
+    email:     "jane.doe@sap.com"
```
IvoGoman, line 52: Could you rename the email addresses to `@example.com`?

uwe-mayer: :100:

IvoGoman (approved):
Thanks 🚀

PR #24, opened 2024-04-02
feat(plugin): rename Plugin & PluginConfig CRDs

IvoGoman:
**DO NOT MERGE** the merge will be done after the CRDs have been migrated

### DIFF
--- a/alerts/charts/templates/am-notification-template-configmap.yaml
+++ b/alerts/charts/templates/am-notification-template-configmap.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: greenhouse-notification-templates
   labels:
-    plugin: alerts
+    pluginDefinition: alerts
     pluginconfig: {{ $.Release.Name }}
     app: prometheus-alertmanager
     alertmanager: {{ include "kube-prometheus-stack.name" . }}
--- a/alerts/charts/templates/am-notification-template-configmap.yaml
+++ b/alerts/charts/templates/am-notification-template-configmap.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: greenhouse-notification-templates
   labels:
-    pluginDefinition: alerts
+    plugin: alerts
     pluginconfig: {{ $.Release.Name }}
     app: prometheus-alertmanager
     alertmanager: {{ include "kube-prometheus-stack.name" . }}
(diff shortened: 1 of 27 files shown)


alerts/charts/templates/am-notification-template-configmap.yaml
```diff
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: greenhouse-notification-templates
   labels:
-    plugin: alerts
+    pluginDefinition: alerts
     pluginconfig: {{ $.Release.Name }}
```
uwe-mayer, line 9:
```suggestion
    plugin: {{ $.Release.Name }}
```

uwe-mayer: Ah ok, we might need to inform plugin owner, if that label changes...

alerts/plugin.yaml
```diff
@@ -51,7 +51,7 @@ spec:
       required: false
       type: list
     - name: alerts.alertmanager.serviceMonitor.additionalLabels
-      description: Additional ServiceMonitor labels. Name of the kube-monitoring PluginConfig to monitor Alertmanager.
+      description: Additional ServiceMonitor labels. Name of the kube-monitoring Plugin to monitor Alertmanager.
       default:
         pluginconfig: kube-monitoring
```
uwe-mayer, line 56: Same here: Plugin owner would need to know, if label changes?

uwe-mayer (approved):

alerts/plugin.yaml
IvoGoman: This only changed the description, but not the label itself. I will revert the description change and leave the defaults/ labels as `pluginconfig: kube-monitoring`, as this is something only used internally in this plugin and can be adjusted later. @richardtief FYI, I believe this can be done after changing the CRDs as these labels are only used internally in the alerts & kube-monitoring plugins, right?

alerts/charts/templates/am-notification-template-configmap.yaml
IvoGoman: good point I will revert this for now. As this is not necessarily connected to the CRD itself but an internal of the chart

alerts/plugin.yaml
richardtief: Yes, they simply label the resources provided with the plugin. Can be adjusted later

PR #33, opened 2024-04-04
fix(reuse) finxing reuse tool findings

kengou:
Trying to fix ruse tool findings


LICENSES /Apache-2.0.txt
IvoGoman, line 1: Looking at the reuse tool documentation, we could also replace this file with a symlink pointing to the licence file in the repository root: https://sap.github.io/fosstars-rating-core/oss_rules_of_play_rating.html  under section
> Why do I need to specify a license in the root folder of my project and as well in the LICENSES folder?

kengou: symlink created

PR #68, opened 2024-04-16
feat(logshipper): add prometheus metrics output

IvoGoman:


logshipper/plugin.yaml
```diff
@@ -7,11 +7,11 @@ metadata:
   name: logshipper
 spec:
   description: Logshipping of container logs and systemd with fluent-bit
-  version: 0.0.1
+  version: 0.0.2
   helmChart:
     name: logshipper
-    repository: oci://ghcr.io/cloudoperators/greenhouse-extensions
-    version: 0.0.1
+    repository: oci://keppel.eu-de-1.cloud.sap/greenhouse-extensions
```
auhlig, line 13: We must use publicly available container registries otherwise this cannot be used for ORA

IvoGoman: Fixed.

auhlig (approved):

PR #73, opened 2024-04-19
fix: security context to allow disk writes

viennaa:
current settings did not allow prometheus to write files. the folder was created correctly under /data but it then failed with missing write permissions

I am happy to start the discussion, whether there is a better way here or not, this however is working, the old settings weren't.


richardtief (approved):
Thanks! If the default values are not reasonable, then we will override them.

PR #87, opened 2024-04-29
feat(node-exporter): nodename label added

richardtief:


kube-monitoring/charts/values.yaml
```diff
@@ -128,6 +128,19 @@ kubeMonitoring:
...
+  ## Configuration for prometheus-node-exporter subchart
+  ##
+  prometheus-node-expoter:
+    enabled: true
+
+    relabelings:
+      - sourceLabels: [__meta_kubernetes_pod_node_name]
+        separator: ;
+        regex: ^(.*)$
+        targetLabel: nodename
```
databus23, line 140: We are currently using the label `node` in all our clusters. So I would suggest we keep it consistency, otherwise dashboards will creak when switching datasources from a greenhouse monitoring stack to one of our existing ones

databus23 (changes requested):

kube-monitoring/charts/values.yaml
databus23: Whatever we use it definitely should match the label that is used in kube-state-metrics so that its easy to create joins between node-exporter and cube-state-metrics timeseries.

richardtief: Fair enough. kube-state-metrics using a label `node`, so I will adopt this.

databus23 (approved):

viennaa (approved):
Thanks

PR #92, opened 2024-05-03
fix(cAdvisor metrics): more useful metric relabeling

richardtief:
ensures one unique metric per container + container label exists


viennaa (approved):
Let's go

PR #98, opened 2024-05-08
back to default context

viennaa:
reverts 755db57


richardtief (changes requested):
Thanks. Could you please increase the plugin version as well?

richardtief (approved):
Thanks!

PR #123, opened 2024-06-03
fix(heureka): Show the general info for each service

hodanoori:
The list of info which should be shown in this phase for each service is updated and are showing the correct general info about each service.


ArtieReus (approved):
lgtm

PR #131, opened 2024-06-07
fix(heureka): make app public, remove stars from peerDependencies and some typos

ArtieReus:
This PR contains:

- Set the MFE public
- Remove stars from the peerDependencies
- Set some peerDependencies version equal to the devDependencies
- Fixes some typos appProps description
- Patch version upgrade


hodanoori (approved):
lgtm

PR #147, opened 2024-06-18
feat(heureka): Impl pagination using juno-ui-comp in services tab

hodanoori:
## Pull Request Details

Pagination is changed in services tab using juno-ui-components instead of using the self customised pagination component.


ArtieReus (approved):
lgtm

PR #164, opened 2024-06-24
fix (heureka): Update the package-lock.json

hodanoori:
## Pull Request Details

Update the package-lock.json to solve the errors (e.g missing darwin pkg error) while building the project locally on mac. 
Apparently some packages are not updated. For this purpose I performed the following commands:

- npm cache clean --force
- npm install

As it is shown in the changes, the `sass` and `graphql-request` as direct packages are updated and the rest are the depended packages.


ArtieReus (dismissed):
lgtm

ArtieReus (changes requested):
would it make sense to upgrade the version here also?

ArtieReus (approved):
lgtm

PR #175, opened 2024-06-26
feat(heureka): Substitute vulnerabilities with  issues tab

hodanoori:
## Pull Request Details

The vulnerabilitiesMatch entity is not relevant anymore and there is a new entity named "issuesMatch" which includes all kind of vulnerabilities. 
The code as well as queries are adjusted to fetch data correctly based on new API changes because of this new entity.


ArtieReus (approved):
lgtm

PR #177, opened 2024-06-27
feat(heureka) : Refactor store implementation to implement filtering functionalities for diferent entities

hodanoori:
## Pull Request Details

Refactor store to have different slices to separate complex filtering relevant implementation from the rest of auth and global store slices. 
In this PR only filtering on services tab is activated now.


heureka/ui/src/App.js
```diff
@@ -27,6 +34,33 @@ const App = (props) => {
   w-full
   `
 
+  useLayoutEffect(() => {
+    // filterLabels are the labels shown in the filter dropdown, enabling users to filter services based on specific criteria. Default is support-group.
+    if (props.filterLabels) setLabels(props.filterLabels)
+
+    // predefined filters config
+    const predefinedFilters = [
```
edda, line 42: it seems like you're not using this array. In line 50 the setter is commented out. If you don't need this better delete it altogether. :)

heureka/ui/src/App.js
```diff
@@ -27,6 +34,33 @@ const App = (props) => {
...
+        name: "support-group",
+        displayName: "support-group",
+        // regex that matches all entities with default filter
+        region: "\bcontainers\b",
+      },
+    ]
+    // setPredefinedFilters(predefinedFilters)
+
+    // initially active predefined filter
+    const initialPredefinedFilter = "support-group"
```
edda, line 53: This is fine for now but you should make this configurable eventually.

heureka/ui/src/App.js
```diff
@@ -27,6 +34,33 @@ const App = (props) => {
...
+    // setActivePredefinedFilter(initialPredefinedFilter)
+
+    // save the apiEndpoint. It is also used outside the alertManager hook
+    setApiEndpoint(props.endpoint)
+  }, [])
+
+  useLayoutEffect(() => {
```
edda, line 60: again the layout effect question. Does this have to be a layouteffect?

heureka/ui/src/components/filters/FilterSelect.jsx
```diff
@@ -46,80 +59,65 @@ const FilterSelect = ({ filters }) => {
...
-          key={resetKey}
         >
-          {filterLabelValues[filterLabel]?.values
-            ?.filter(
-              (value) =>
-                // filter out already active values for this label
-                !activeFilters[filterLabel]?.includes(value)
-            )
+          {filterLabelValues[filterLabel]
+            ?.filter((value) => !activeFilters[filterLabel]?.includes(value))
```
edda, line 97: I would add the comment about what this does back in. It's a bit hard to read the long function chain without some context.

heureka/ui/src/components/filters/FilterSelect.jsx
```diff
@@ -46,80 +59,65 @@ const FilterSelect = ({ filters }) => {
...
-      {/* <SearchInput
-        placeholder="search term or regular expression"
-        className="w-96 ml-auto"
-        value={searchTerm || ""}
-        onSearch={(value) => setSearchTerm(value)}
-        onClear={() => setSearchTerm(null)}
-        onChange={(value) => handleSearchChange(value)}
-      /> */}
+      )}
+      {entityName === "IssueMatches" && (
```
edda, line 111: Why the search input only for `IssueMatches`?

heureka/ui/src/components/filters/Filters.jsx
```diff
@@ -18,28 +20,39 @@ const filtersStyles = `
...
+      const field = data[queryKey][key]
+      return {
+        label: field.filterName, // Collecting filterName as filterLabel
+        values: field.values, // Collecting values as filterValues
+      }
+    })
+  }, [data, queryKey])
+
+  useEffect(() => {
+    if (filters.length > 0) {
```
edda, line 45: This could use a comment. What are you doing here and why?

heureka/ui/src/components/filters/Filters.jsx
```diff
@@ -18,28 +20,39 @@ const filtersStyles = `
...
-    return data?.__type?.inputFields?.map((field) => ({
-      label: field?.name,
-      kind: field?.type?.kind,
-      ofType: field?.type?.ofType?.name,
-      enumValues: field?.type?.ofType?.enumValues,
-    }))
-  }, [data])
+    if (!data || !data[queryKey]) return []
+
+    return Object.keys(data[queryKey]).map((key) => {
```
edda, line 35: This could use a comment. What is happening here?

heureka/ui/src/components/issues/IssuesList.jsx
```diff
@@ -18,12 +18,15 @@ const IssuesList = ({ items, isLoading }) => {
   return (
     <>
       {/* clickableTable Table allow changes the background by css when hovering or active*/}
-      <DataGrid columns={10} className="clickableTable">
+      <DataGrid
+        gridColumnTemplate="2fr 2fr 4fr 2fr 2fr 2fr 2fr 2fr 2fr 2fr"
```
edda, line 22: is it really necessary to manually set the third column to double the size? If it has more content than the other columns then the auto layout should pretty much take care of it already.

heureka/ui/src/components/services/ServicesList.jsx
```diff
@@ -27,8 +27,10 @@ const ServicesList = ({ items, isLoading }) => {
           <DataGridHeadCell>Service</DataGridHeadCell>
           <DataGridHeadCell>Owners</DataGridHeadCell>
           <DataGridHeadCell>Support Groups</DataGridHeadCell>
-          <DataGridHeadCell>Component Instances Count</DataGridHeadCell>
-          <DataGridHeadCell>Total Vulnerability Matches</DataGridHeadCell>
+          <DataGridHeadCell>
+            Total Number of Component Instances
```
edda, line 31: In general I would advise not to use these super long Headings. It makes the table render weirdly if the content in the cells is a lot smaller than the heading for the column. If the long text is essential then my suggestion is to use a short version of the text as the heading and add a tooltip to it with the long form version.

heureka/ui/src/components/services/ServicesList.jsx
```diff
@@ -27,8 +27,10 @@ const ServicesList = ({ items, isLoading }) => {
...
+          </DataGridHeadCell>
+          <DataGridHeadCell>Total Number of Issue Matches</DataGridHeadCell>
```
edda, line 33: see above

heureka/ui/src/lib/slices/createFiltersSlice.js
```diff
@@ -0,0 +1,227 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { produce } from "immer"
+
+const initialFiltersState = {
+  labels: {}, // Labels for each entity: { entityName: ["label1", "label2", ...] }
+  activeFilters: {}, // Active filters for each entity: { entityName: { label1: [value1], label2: [value2_1, value2_2], ... } }
```
edda, line 10: I am not a fan of deeply nested state out of previous painful experience. It makes it easier to make mistakes when updating values somewhere deep in the hierarchy. The `immer` library helps with this but if you forget to use it even once you can accidentally overwrite a huge chunk of the structure. 

I would strongly suggest reworking this (maybe not now but in the future) for the state value to be as flat as possible. I.e. rather than having one state `activeFilters` that contains the active filters for all possible entities to instead have `entity1ActiveFilters`, `entity2ActiveFilters`, etc. This will result in a bit more boilerplate but it will be much easier to understand and read and also to debug when something goes wrong.

The actions should still be doable in such a way that you don't have to copy the code.

The same goes for `filterLabelValues`, `predefinedFilters`, `activePredefinedFilters`, ...

heureka/ui/src/lib/slices/createFiltersSlice.js
```diff
@@ -0,0 +1,227 @@
...
+            state.filters.filterLabelValues[entityName][filterLabel] = {
+              isLoading: true,
+            }
+          }),
+          false,
+          "filters.loadFilterLabelValues.isLoading"
+        )
+
+        // Simulate a data fetch and update the state
+        set(
```
edda, line 183: what does this do exactly? Which part is simulated?

heureka/ui/src/lib/slices/createGlobalsSlice.js
```diff
@@ -0,0 +1,120 @@
...
+ * SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { produce } from "immer"
+import constants from "../../components/shared/constants"
+
+const createGlobalsSlice = (set, get, options) => ({
+  globals: {
+    embedded: false,
```
edda, line 11: all of these state values need a comment to explain what they are. Not all are self-explanatory.

heureka/ui/src/helpers.js
```diff
@@ -20,6 +20,14 @@ export const parseError = (error) => {
   return errMsg
 }
 
+export const formatLabel = (label) => {
```
edda, line 23: It looks like this does a very similar thing to the `humanizeString` function you already have in utils. Is this helper really necessary?

heureka/ui/src/lib/slices/createAuthDataSlice.js
```diff
@@ -0,0 +1,71 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const AUTH_ACTIONS = {
+  SIGN_ON: "signOn",
+  SIGN_OUT: "signOut",
+}
+
+const createAuthDataSlice = (set, get) => ({
+  auth: {
```
edda, line 12: This goes for all store slices: Please add a ticket for yourself to unit test the slices once you are happy with how the store is designed.

edda (changes requested):

heureka/ui/package.json
```diff
@@ -73,6 +72,11 @@
       "value": "false",
       "type": "optional",
       "description": "Set to true if app is to be embedded in another existing app or page.  If set to true the app won't render a page header/footer and instead render only the content. The default value is false."
+    },
+    "filterLabels": {
```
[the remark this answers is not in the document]
hodanoori: No, this is not nedded, as you said the filter labels are read from BE.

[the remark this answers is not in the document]
hodanoori: No, this is not configurable in FE. I deleted it.

heureka/ui/src/App.js
hodanoori: I will set the predefined filter of support group in a separate PR.

heureka/ui/src/App.js
```diff
@@ -27,6 +34,33 @@ const App = (props) => {
   w-full
   `
 
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
hodanoori: I removed this part of setting default filter value to do it in a separate PR.

heureka/ui/src/lib/slices/createFiltersSlice.js
hodanoori: Sure, I restructure this soon in my next PRs.

heureka/ui/src/lib/slices/createFiltersSlice.js
hodanoori: Left over from the first impl of filter for services tab. I deleted it.

heureka/ui/src/helpers.js
hodanoori: there is a slight difference that in formatLabel, all vocabularies are started with capital case. But it is not that important, I deleted this function and used the humanize string to keep the consistency in greenhouse as the Alerts filter labels are using this formatter as well.

heureka/ui/src/lib/slices/createAuthDataSlice.js
hodanoori: Sure, I will

heureka/ui/src/components/issues/IssuesList.jsx
edda:
```suggestion
        columns={10}
```

edda (changes requested):

edda:
There are also a bunch of unresolved comments in the middle. Mostly about adding comments to explain what certain things do.

heureka/ui/src/App.js
hodanoori: I think so, as I explained in the other section.

hodanoori:
> There are also a bunch of unresolved comments in the middle. Mostly about adding comments to explain what certain things do.

Right, sorry! I overseen them. I resolved them all, hopefully!

edda (approved):

PR #186, opened 2024-06-28
fix(alerts/ui): TopNav is now also shown in the embedded mode

TilmanHaupt:
## Pull Request Details

The TopNavigation of alerts is now also shown in the embedded mode.

## Issues Fixed

fixes: #185


edda (approved):
Temporary fix until embedded mode is fixed in `AppShell`.

PR #192, opened 2024-07-04
feat(thanos): enabling thanos sidecar functionality

viennaa:
## Pull Request Details

This will enable thanos sidecar functionality on kube-monitoring side.

### DIFF
--- a/kube-monitoring/charts/Chart.yaml
+++ b/kube-monitoring/charts/Chart.yaml
@@ -8,7 +8,7 @@ maintainers:
 name: kube-monitoring
 sources:
   - https://github.com/cloudoperators/greenhouse-extensions
-version: 0.8.2
+version: 0.8.3
 # prometheus-operator app version
 appVersion: v0.73.2
 keywords:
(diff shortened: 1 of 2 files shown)


kube-monitoring/charts/Chart.yaml
```diff
@@ -8,7 +8,7 @@ maintainers:
 name: kube-monitoring
 sources:
   - https://github.com/cloudoperators/greenhouse-extensions
-version: 0.8.2
+version: 0.8.3
```
richardtief, line 11: That's not necessary, is it? There is no actual change in the chart and it is not reflected in the plugin below. 
```suggestion
Version: 0.8.2
```

richardtief (changes requested):
Thanks. Just remove the chart bump. The plugin bump is valid

kube-monitoring/charts/Chart.yaml
viennaa: True, was due to the first intended change. Pushing again to keep `version` with lower `v`

richardtief (approved):

PR #204, opened 2024-07-10
[opentelemetry-operator] add CRDs flags to values, rm wrong license, …

timojohlo:
---

## Pull Request Details

Currently the OpenTelemetry Operator only allows to install all CRDs or no CRD at all. With the availability of the OpenTelemetry Plugin in Greenhouse, users should be able to opt-in/opt-out on each CRD individually, so they can decide what exporter to use, what collector to use etc.

## Breaking Changes

None

## Issues Fixed

None

## Other Relevant Information

Adapting changes after #202

### DIFF
--- a/opentelemetry-operator/chart/Chart.yaml
+++ b/opentelemetry-operator/chart/Chart.yaml
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.0.5
+version: 0.0.6
 description: OpenTelemetry Operator Helm chart for Kubernetes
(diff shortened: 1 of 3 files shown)


Kuckkuck (approved):

richardtief:
Thank you. You are using a mixed approach for installing CRDs on the cluster. Either you use the `crds` folder and leave the installation to Helm or you create a separate chart to install them. Please read https://helm.sh/docs/chart_best_practices/custom_resource_definitions 

Otherwise, we want to install them all when the plugin is instantiated. It wouldn't hurt to have them available, right?

richardtief:
There is also a ServieMonitor in the `/crds` folder, which should be located in `/templates`. ServiceMonitor CRD is installed by the kube-monitoring plugin. In this case, this plugin is a dependency. 🙃

PR #219, opened 2024-07-12
[opentelemetry-operator] template image, bump chart version

timojohlo:
## Pull Request Details

Template OpenTelemetry Collector Image so it can be passed through the PluginDefinition.

### DIFF
--- a/opentelemetry-operator/chart/templates/logs-collector.yaml
+++ b/opentelemetry-operator/chart/templates/logs-collector.yaml
@@ -8,7 +8,7 @@ kind: OpenTelemetryCollector
 metadata:
   name: logs 
 spec:
-  image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.104.0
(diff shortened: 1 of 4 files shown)


IvoGoman (approved):

opentelemetry-operator/chart/templates/logs-collector.yaml
```diff
@@ -8,7 +8,7 @@ kind: OpenTelemetryCollector
 metadata:
   name: logs 
 spec:
-  image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.104.0
+  image: {{ index .Values "opentelemetry-operator" "manager" "collectorImage" "repository" }}:{{ index .Values "opentelemetry-operator" "manager" "collectorImage" "tag" }}
```
richardtief, line 11: Note: You do not always have to use the index function because there is a `-` in the name of the dependent chart. You can also define an alias for it. https://github.com/cloudoperators/greenhouse-extensions/blob/c31b80ca5750d21b18842d7614d925ea51ad7224/kube-monitoring/charts/Chart.yaml#L20

PR #224, opened 2024-07-16
feat(alerts/ui): improvements for filters and details 

### LINKED ISSUE #200: feat(alerts/ui): Make active filters toggleable
Make filters toggleable

- [x] Ensure that filters are activating and deactivating with a click on the pill
- [x] Ensure they are still closable
- [x] Ensure that that filters are in the same order
- [x] Add to urlState 
- [x] Add TopNav to appshell as prop

TilmanHaupt:
# Pull Request Details

- Filtered Pills can now be de- and reactivated
- In Alerts Details Silences can now be expired or recreated

Active filter pills can now be clicked to deactivate them and clicked again to activate them.
closes #200 #274 #344

### DIFF
--- a/alerts/ui/src/components/filters/FilterPills.jsx
+++ b/alerts/ui/src/components/filters/FilterPills.jsx
@@ -6,30 +6,69 @@
 import React from "react"
 
 import { Pill, Stack } from "@cloudoperators/juno-ui-components"
-import {
-  useActiveFilters,
-  useFilterActions,
-  usePausedFilters,
-} from "../../hooks/useAppStore"
+import { useActiveFilters, useFilterActions } from "../../hooks/useAppStore"
 
 const FilterPills = () => {
   const activeFilters = useActiveFilters()
-  const pausedFilter = usePausedFilters()
-  const {
-    removeActiveFilter,
-    addPausedFilter,
-    addActiveFilter,
-    removePausedFilter,
-  } = useFilterActions()
+  const [filters, setFilters] = useState({})
+  const { removeActiveFilter } = useFilterActions()
+
+  // useEffect Hook zur Aktualisierung der Filter
+  useEffect(() => {
+    setFilters((prevFilters) => {
+      const newFilters = { ...prevFilters }
+
+      for (let key in newFilters) {
+        if (newFilters.hasOwnProperty(key)) {
+          if (activeFilters[key]) {
+            const activeSet = new Set(activeFilters[key])
+
+            // Behalte inaktive Elemente und aktualisiere aktive
+            newFilters[key] = newFilters[key].map((item) => {
+              if (activeSet.has(item.value)) {
+                return { ...item, active: true }
+              }
+              return item // Behalte inaktive Elemente unverändert
+            })
+
+            // Füge neue aktive Elemente hinzu
+            activeFilters[key].forEach((value) => {
+              if (!newFilters[key].some((item) => item.value === value)) {
+                newFilters[key].push({ value, active: true })
+              }
+            })
+          }
+        }
+      }
+
+      // Hinzufügen neuer Kategorien aus activeFilters
+      for (let key in activeFilters) {
+        if (!newFilters.hasOwnProperty(key)) {
+          newFilters[key] = activeFilters[key].map((value) => ({
+            value,
+            active: true,
+          }))
+        }
+      }
+
+      return newFilters
+    })
+    console.log(filters)
+  }, [activeFilters])
 
   const pauseFilter = (key, value) => {
-    removeActiveFilter(key, value)
-    addPausedFilter(key, value)
-  }
+    setFilters((prevFilters) => {
+      const newFilters = { ...prevFilters }
+      if (newFilters[key]) {
+        newFilters[key] = newFilters[key].map((item) =>
+          item.value === value ? { ...item, active: false } : item
+        )
+      }
+      return newFilters
+    })
 
-  const activateFilter = (key, value) => {
-    removePausedFilter(key, value)
-    addActiveFilter(key, value)
+    // Optional: Aktualisiere auch activeFilters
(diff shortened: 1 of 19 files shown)


alerts/ui/src/components/filters/FilterPills.jsx
```diff
@@ -22,6 +42,21 @@ const FilterPills = () => {
             closeable
             onClose={() => removeActiveFilter(key, value)}
             key={`${key}:${value}`}
+            onClick={() => pauseFilter(key, value)}
+          />
+        ))
+      })}
+      {Object.entries(pausedFilter).map(([key, values]) => {
+        console.log("sdfsdf", values)
```
edda, line 50: please remove `console.log` :)

alerts/ui/src/components/filters/FilterPills.jsx
```diff
@@ -22,6 +42,21 @@ const FilterPills = () => {
...
+        return values.map((value) => (
+          <Pill
+            className="bg-theme-background-lvl-4"
```
edda, line 53: Instead of adjusting the bg color, try changing the opacity of the paused filters, e.b. `opacity-70`

alerts/ui/src/components/filters/FilterPills.jsx
```diff
@@ -22,6 +42,21 @@ const FilterPills = () => {
             closeable
             onClose={() => removeActiveFilter(key, value)}
             key={`${key}:${value}`}
+            onClick={() => pauseFilter(key, value)}
+          />
+        ))
+      })}
+      {Object.entries(pausedFilter).map(([key, values]) => {
```
edda, line 49: If I read this correctly you're moving all paused filters to the end of the Pill Stack. That can lead to undesirable UX because the Pill someone clicked to pause it gets yanked away from its original position. It's much better to leave it in place and only change its state.

edda (changes requested):

alerts/ui/src/components/filters/FilterPills.jsx
TilmanHaupt: done

alerts/ui/src/components/filters/FilterPills.jsx
```diff
@@ -3,27 +3,141 @@
...
+          }))
+        }
+      }
+
+      return newFilters
+    }
+    setFilterPills(pills(filters))
+  }, [activeFilters])
+
+  const pauseFilter = (key, value) => {
```
edda, line 61: Why are this and the following functions that all modify the filters defined here and not as actions in the store?

alerts/ui/src/components/filters/FilterPills.jsx
```diff
@@ -3,27 +3,141 @@
...
 
 const FilterPills = () => {
   const activeFilters = useActiveFilters()
-  const { removeActiveFilter } = useFilterActions()
+  const filters = useFilterPills()
+  const { removeActiveFilter, addActiveFilter, setFilterPills } =
+    useFilterActions()
+  // useEffect Hook zur Aktualisierung der Filter
+  useEffect(() => {
+    const pills = (prevFilters) => {
```
edda, line 22: Creating this new merged object seems very complicated and potentially error prone to me. 

An easier solution might be store the two objects for active and paused filters (where paused filters are a subset of active filters). Then when you render the filter pills you do it by iterating over the active filters, and for each pill you check whether that key:value combination is also present in pausedFilters and then render accordingly. That would make it much easier to understand what is happening, and it doesn't require you to run through the list twice.

Let's have a chat about this.

edda (changes requested):

alerts/ui/src/lib/createFiltersSlice.js
```diff
@@ -8,10 +8,12 @@ import { produce } from "immer"
 const initialFiltersState = {
   labels: ["status"], // labels to be used for filtering: [ "label1", "label2", "label3"]. Default is status which is enriched by the worker
   activeFilters: {}, // for each active filter key list the selected values: {key1: [value1], key2: [value2_1, value2_2], ...}
+  pausedFilters: {}, // for each active filter key list the selected values: {key1: [value1], key2: [value2_1, value2_2], ...}
```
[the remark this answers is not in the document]
TilmanHaupt: This is old code from another Approach. > deleted

alerts/ui/src/components/filters/FilterPills.jsx
TilmanHaupt: Thats not possible because of the architecture of Alerts. Because the "active" label was not existent before the app handels all filters in this Array as active. This change would lead to a refactor of various components.

alerts/ui/src/components/filters/FilterPills.jsx
TilmanHaupt: Because they are just important for this component and would lead to a more filled Store. FilterPills are just useful in the store for deeplink Implantation which is not done now.

alerts/ui/src/components/filters/FilterPills.jsx
edda: hmm, I'm not sure I understand the problem you describe. But I've probably not explained well enough what I meant. My proposed solution was something like the following:
1. add a `pauseFilter/unpauseFilter` action to the `filter` store slice that works similar to the `removeActiveFilter/addActiveFilter` actions in that it modifies the pausedFilters state and then calls the `filterItems` action of the `alerts` slice afterwards. 
2. In the `filterItems` action we must then modify the section that checks whether the item is being filtered out by an active filter to also take paused filters into account
3. When rendering the list of filter pills you also need to add the check to figure out if the filter is paused as discussed above.

edda (changes requested):

edda (changes requested):
I like it!

edda (approved):

PR #237, opened 2024-07-23
Add workflow for building and publishing integration test Docker image

ibakshay:
This pull request adds a new workflow for building and publishing an integration test Docker image. The workflow is triggered when there is a change in `Docker.integration-test` file and when a tag matching the pattern "v*.*.*" is pushed. It builds the Docker image using the Dockerfile.integration-test file and pushes it to the registry. Additionally, a vulnerability scan step is included to scan the image for vulnerabilities using Trivy and upload the scan results to the GHAS. 

I have already tested this workflow in my [personal github repository](https://github.com/ibakshay/k8s-infra/pkgs/container/greenhouse-extensions-integration-test).

### DIFF
--- a/.github/workflows/integration-test-build-publish.yaml
+++ b/.github/workflows/integration-test-build-publish.yaml
@@ -17,7 +17,7 @@ env:
 jobs:
   build-and-push:
     name: Build and push Docker image
-    runs-on: ubuntu-latest
+    runs-on: [default]
     permissions:
       contents: read
       packages: write
@@ -85,7 +85,7 @@ jobs:
       security-events: write
     name: Vulnerability Scan
     needs: build-and-push
-    runs-on: ubuntu-latest
+    runs-on: [default]
     steps:
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@0.20.0
--- a/.github/workflows/integration-test-build-publish.yaml
+++ b/.github/workflows/integration-test-build-publish.yaml
@@ -88,7 +88,7 @@ jobs:
     runs-on: [default]
     steps:
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@0.20.0
+        uses: aquasecurity/trivy-action@0.24.0
         with:
           image-ref: ${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}:${{ github.sha }}
           ignore-unfixed: true
--- a/.github/workflows/integration-test-build-publish.yaml
+++ b/.github/workflows/integration-test-build-publish.yaml
@@ -1,0 +1,103 @@
+name: Build and Publish Integration Test Docker Image
+
+on:
+  workflow_dispatch: {}
+  push:
+    branches:
+      - main
+    paths:
+      - Dockerfile.integration-test
+    tags:
+      - v*.*.*
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: greenhouse-extensions-integration-test
+
+jobs:
+  build-and-push:
+    name: Build and push Docker image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Install cosign
+        if: github.event_name != 'pull_request'
+        uses: sigstore/cosign-installer@v3.5.0
+        with:
+          cosign-release: "v2.2.3"
+      - name: Set up QEMU # For multi-platform builds
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx # For multi-platform builds
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver-opts: |
+            image=moby/buildkit:latest
+      - name: Log into registry ${{ env.REGISTRY }}
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=semver,pattern={{version}},prefix=v
+            type=semver,pattern={{major}}.{{minor}},prefix=v
+            type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/v0.') }},prefix=v
+            type=ref,event=branch
+            type=ref,event=tag
+            type=ref,event=pr
+            type=schedule
+            type=raw,value=${{ github.sha }}
+            type=sha,enable=true,format=short,prefix=
+      - name: Build and push Docker image
+        id: build-and-push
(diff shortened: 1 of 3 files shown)


.github/workflows/integration-test-build-publish.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - v*.*.*
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: greenhouse-extensions-integration-test
+
+jobs:
+  build-and-push:
+    name: Build and push Docker image
+    runs-on: ubuntu-latest
```
kengou, line 20: Please use the GitHub organization provided runner: `runs-on: [ default ]`

.github/workflows/integration-test-build-publish.yaml
```diff
@@ -0,0 +1,103 @@
...
+        run: echo "${TAGS}" | xargs -I {} cosign sign --yes {}@${DIGEST}
+
+  vulnerability-scan:
+    permissions:
+      contents: read
+      packages: read
+      security-events: write
+    name: Vulnerability Scan
+    needs: build-and-push
+    runs-on: ubuntu-latest
```
kengou, line 88: Please change to the organisation provided runner `[default]`

.github/workflows/integration-test-build-publish.yaml
```diff
@@ -0,0 +1,103 @@
...
+        id: build-and-push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: Dockerfile.integration-test
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
kengou, line 72: We had issues using the `gha` cache during build in the past.

Dockerfile.integration-test
```diff
@@ -0,0 +1,14 @@
+FROM bats/bats:latest
+
+ARG KUBECTL_VERSION=v1.29.2
+
+# Add packages
+RUN apk --no-cache add \
+    curl \
+    git \
+    bash
```
kengou, line 9: I would run an upgrade on all installed packages to fix potential CVEs in advance
```suggestion
RUN apk upgrade --no-cache --no-progress \
&& apk --no-cache --no progress add curl git bash \
&& apk del --no-cache --no-progress apk-tools alpine-keys
```

Dockerfile.integration-test
```diff
@@ -0,0 +1,14 @@
+FROM bats/bats:latest
```
kengou, line 1: I would like to avoid the latest tag and just use an exact versioning what we update
```suggestion
FROM bats/bats:1.11.0
```

Dockerfile.integration-test
```diff
@@ -0,0 +1,14 @@
...
+
+ARG KUBECTL_VERSION=v1.29.2
```
kengou, line 3: In greenhouse we use 1.30.2 version. would be nice to keep the versions aligned everywhere
```suggestion
ARG KUBECTL_VERSION=v1.30.2
```

kengou (changes requested):

.github/workflows/integration-test-build-publish.yaml
ibakshay: > We had issues using the `gha` cache during build in the past.

okay. I will keep an eye on it.

.github/workflows/integration-test-build-publish.yaml
ibakshay: thank you. I have fixed in https://github.com/cloudoperators/greenhouse-extensions/pull/237/commits/924cf43397727fe509b9c81e936192123c6d484f

Dockerfile.integration-test
ibakshay: I have fixed in https://github.com/cloudoperators/greenhouse-extensions/pull/237/commits/924cf43397727fe509b9c81e936192123c6d484f commit.

Dockerfile.integration-test
ibakshay: I have fixed in https://github.com/cloudoperators/greenhouse-extensions/pull/237/commits/924cf43397727fe509b9c81e936192123c6d484f commit.

Dockerfile.integration-test
ibakshay: I have fixed in https://github.com/cloudoperators/greenhouse-extensions/pull/237/commits/924cf43397727fe509b9c81e936192123c6d484f commit.

kengou (approved):

.github/workflows/integration-test-build-publish.yaml
```diff
@@ -0,0 +1,103 @@
...
+          platforms: linux/amd64,linux/arm64
```
IvoGoman, line 73: There are 2 Platforms defined here, but the Dockerfile explicitly downloads a linux/amd64 binary. 
https://github.com/cloudoperators/greenhouse-extensions/pull/237/files#diff-c488c0b59bfe42850bda02e59aff3bd6ee6fc6e2a1a52a8841ca3f0d100b631eR11

kengou: Also then the `FROM bats/bats:1.11.0` needs to be changed. 
In greenhouse we use  like this `FROM --platform=${BUILDPLATFORM:-linux/amd64} bats/bats:1.11.0`

.github/workflows/integration-test-build-publish.yaml
```diff
@@ -0,0 +1,103 @@
...
+    permissions:
+      contents: read
+      packages: read
+      security-events: write
+    name: Vulnerability Scan
+    needs: build-and-push
+    runs-on: [default]
+    steps:
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@0.20.0
```
kengou, line 91: Did you have any special reason not to use the latest `trivy-action`?
```suggestion
        uses: aquasecurity/trivy-action@0.24.0
```

.github/workflows/integration-test-build-publish.yaml
kengou: We may can use: `https://storage.googleapis.com/kubernetes-release/release/$KUBECTL_VERSION/bin/${BUILDPLATFORM:-linux/amd64}/kubectl`

Dockerfile.integration-test
```diff
@@ -0,0 +1,13 @@
...
+
+# Add packages
+RUN apk upgrade --no-cache --no-progress \
+&& apk --no-cache --no-progress add curl git bash \
+&& apk del --no-cache --no-progress apk-tools alpine-keys
+
+# Install kubectl binary
+RUN curl -LO "https://storage.googleapis.com/kubernetes-release/release/$KUBECTL_VERSION/bin/linux/amd64/kubectl" && \
+	chmod +x kubectl && \
+	mv kubectl /usr/local/bin/
```
kengou, line 13:
```suggestion
	mv kubectl /usr/local/bin/
	
```

.github/workflows/integration-test-build-publish.yaml
ibakshay: Thank you for looking into it. I used the version mentioned in [trivy-action's README](https://github.com/aquasecurity/trivy-action/blob/master/README.md). Looks like the version is not update in their README. I have also [created a PR in their repo](https://github.com/aquasecurity/trivy-action/pull/378) to update the versions in their README.

richardtief (approved):
All good. Let's ship it

PR #246, opened 2024-07-25
[opentelemetry-operator] update CRDs

timojohlo:
## Pull Request Details

-  add appVersion
- update CRDs with correct source: https://github.com/open-telemetry/opentelemetry-operator/tree/main/config/crd/bases
(instead of: https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-operator/conf/crds)

### DIFF
--- a/opentelemetry-operator/chart/Chart.yaml
+++ b/opentelemetry-operator/chart/Chart.yaml
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.1.3
+version: 0.1.4
 description: OpenTelemetry Operator Helm chart for Kubernetes
 icon: https://raw.githubusercontent.com/cncf/artwork/a718fa97fffec1b9fd14147682e9e3ac0c8817cb/projects/opentelemetry/icon/color/opentelemetry-icon-color.png
 type: application
--- a/opentelemetry-operator/chart/Chart.yaml
+++ b/opentelemetry-operator/chart/Chart.yaml
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 apiVersion: v2
-appVersion: 0.104.0
+appVersion: v0.104.0
 name: opentelemetry-operator
 version: 0.1.4
 description: OpenTelemetry Operator Helm chart for Kubernetes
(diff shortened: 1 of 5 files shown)


opentelemetry-operator/chart/crds/crd-opentelemetrycollector.yaml
```diff
@@ -21,7 +21,6 @@ spec:
           namespace: {{ .Release.Namespace }}
           path: /convert
           port: {{ .Values.admissionWebhooks.servicePort }}
-  {{ if .caBundle }}{{ cat "caBundle:" .caBundle | indent 8 }}{{ end }}
```
viennaa, line 24: Why is this not needed anymore?

timojohlo: We are not using the caBundle right now and this is currently hindering helm to correctly render the chart.

viennaa: Is this planned to be used in the future? 

If so, add an empty directive in your values and the `if` will evaluate to false.

```yaml
opentelemetry-operator
   [...]
   caBundle:
   [...]
```

timojohlo: Not sure, but I agree with you that I make sense to make this optional. The reason I removed it was that the line
`  {{ if .caBundle }}{{ cat "caBundle:" .caBundle | indent 8 }}{{ end }}
`
is afaik not valid in helm and the function `cat` throws an error when installing the chart.

timojohlo: the source for the CRDs were wrong. The correct source for CRDs is https://github.com/open-telemetry/opentelemetry-operator/tree/main/config/crd/bases

opentelemetry-operator/chart/Chart.yaml
```diff
@@ -2,8 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 apiVersion: v2
+appVersion: 0.104.0
```
viennaa, line 5:
```suggestion
appVersion: v0.104.0
```

I'd name it exactly like the version upstream.

viennaa (changes requested):

viennaa (approved):
Thanks, let's go :)

PR #268, opened 2024-07-30
chore(alerts/ui): Update all dependencies from asset server to public npm

### LINKED ISSUE #260: refactor(alerts/ui): Update all dependencies from internal asset server to public npm for alerts
Refactor components, so that they are loaded from npm and not from the asset server

TilmanHaupt:
# Pull Request Details

Update all dependencies from internal asset server to public npm for the alert plugin

## Issues Fixed

#260


andypf (approved):
LGTM

PR #272, opened 2024-07-30
chore(alerts|kube-monitoring): managed prometheus-operator and CRDs

richardtief:
Installation of the CustomResourceDefinitions and the prometheus operator is a unique task for a cluster. We have two plugins that depend on it and therefore need to be able to bring the basis for their installation themselves. Therefore we decided to use the functionality of the kube-prometheus-stack to manage CRDs and turn it on by default. This makes chart testing considerably easier.

# Submit a pull request

is true:

5. All tests pass.

---

Replace any ":question:" below with information about your pull request.

## Pull Request Details

:question:

## Breaking Changes

:question:

## Issues Fixed

1. :question:

## Other Relevant Information

:question:


alerts/plugindefinition.yaml
```diff
@@ -38,6 +38,14 @@ spec:
       required: false
       type: list
 
+    - name: alerts.prometheusOperator.enabled
+      description: Deploy Prometheus Operator if kube-monitoring has not already installed it once.
+      required: false
+      type: bool
+    - name: alerts.crds.enabled
+      description: Install Prometheus Operator CRDs if kube-monitoring has not already installed them.
```
ibakshay, line 46: Do we need to set a default value for `alerts.crds.enabled`?

richardtief: Yes. We enable it by default in the chart values, but disable it by default for the plugin as we intentionally want kube-monitoring to manage the operator and the CRDs. These options only highlight the possibility for the plugin to act independently

alerts/plugindefinition.yaml
```diff
@@ -38,6 +38,14 @@ spec:
...
+      required: false
+      type: bool
```
ibakshay, line 48: >Yes. We enable it by default in the chart values, but disable it by default for the plugin as we intentionally want >kube-monitoring to manage the operator and the CRDs. These options only highlight the possibility for the >plugin to act independently

@richardtief, Thank you for the explanation. Do we need to set this default value to `false` explicitly like below then? 
Or, Am I missing something? 
```suggestion
      required: false
      default: false
      type: bool
```

alerts/plugindefinition.yaml
```diff
@@ -38,6 +38,15 @@ spec:
       required: false
       type: list
 
+    - name: alerts.prometheusOperator.enabled
+      description: Deploy Prometheus Operator if kube-monitoring has not already installed it once.
+      required: false
+      type: bool
```
ibakshay, line 44:
```suggestion
      required: false
      default: false
      type: bool
```

ibakshay (approved):
LGTM! 🚀

viennaa (commented):
What about updating the documentation of kube-monitoring and alerts to reflect this change? Right now kube-monitoring is stating to deploy the operator and having alerts first is now a hard dependency, no?

richardtief:
No, the behaviour should remain the same. We are just enabling the alerts plugin to operate independently of the kube-monitoring plugin. The documentation thus remains valid - but we should still do a review of the documentation, fair point.

PR #281, opened 2024-08-07
feat(heureka): Add msg handling per tab and mem tabs to prevent unnecessary re-rendering

hodanoori:
## Pull Request Details

- Add messages provider per tab to be able to show errors per tab separately.
- Since all tabs which are using tabNavigationItem juno component are re-rendered after clicking on each tab, have to memorize each tab panel info including messages.
- Update services query based on API changes

### DIFF
--- a/heureka/ui/src/components/components/ComponentsListController.jsx
+++ b/heureka/ui/src/components/components/ComponentsListController.jsx
@@ -41,7 +41,7 @@ const ComponentsListController = () => {
   }, [data])
 
   useEffect(() => {
-    if (!error) resetMessages()
+    if (!error) return resetMessages()
     addMessage({
       variant: "error",
       text: parseError(error),
@@ -87,9 +87,6 @@ const ComponentsListController = () => {
 
   return (
     <>
-      <Container py>
-        <Messages />
-      </Container>
       <Container py>
         <ComponentsList components={components} isLoading={isLoading} />
       </Container>
--- a/heureka/ui/src/components/components/ComponentsListController.jsx
+++ b/heureka/ui/src/components/components/ComponentsListController.jsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React, { useMemo, useState } from "react"
+import React, { useEffect, useMemo, useState } from "react"
 import { useQuery } from "@tanstack/react-query"
 import {
   useQueryClientFnReady,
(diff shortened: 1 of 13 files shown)


heureka/ui/src/components/components/ComponentsListController.jsx
```diff
@@ -73,22 +87,30 @@ const ComponentsListController = () => {
...
-          onPressNext={onPressNext}
-          onPressPrevious={onPressPrevious}
-          onKeyPress={onKeyPress}
-          onSelectChange={onPaginationChanged}
-          pages={totalPages}
-          variant="input"
-        />
-      </Stack>
+      {!!error ? (
+        <Container py>
```
ArtieReus, line 91: You don't have to render the messages container just when error happens. This is something that the library does for you. Just render <Messages /> always and it will display errors if needed.

heureka/ui/src/components/components/ComponentsListController.jsx
```diff
@@ -34,6 +40,14 @@ const ComponentsListController = () => {
     return data?.Components?.edges
   }, [data])
 
+  useEffect(() => {
+    if (!error) return
```
ArtieReus, line 44: You can reset hier the errors if no erros happen.

heureka/ui/src/components/services/ServicesListController.jsx
```diff
@@ -40,7 +41,10 @@ const ServicesListController = () => {
 
   useEffect(() => {
     if (!error) return
```
ArtieReus, line 43: same behaviour as the the other component.

heureka/ui/src/components/services/ServicesListController.jsx
```diff
@@ -85,22 +89,30 @@ const ServicesListController = () => {
...
-          isLastPage={currentPage === totalPages}
-          onPressNext={onPressNext}
-          onPressPrevious={onPressPrevious}
-          onKeyPress={onKeyPress}
-          onSelectChange={onPaginationChanged}
-          pages={totalPages}
-          variant="input"
-        />
-      </Stack>
+      {!!error ? (
```
ArtieReus, line 92: and hier also render always the messages.

ArtieReus (changes requested):

heureka/ui/src/components/components/ComponentsListController.jsx
hodanoori: I tried this, but as there are diferent tabs, without this check, on each tab at least one error message is displayed! therefore I kept it here.

heureka/ui/src/components/services/ServicesListController.jsx
```diff
@@ -82,6 +86,9 @@ const ServicesListController = () => {
       onPaginationChanged(parseInt(oKey.currentTarget.value))
     }
   }
+  useLayoutEffect(() => {
```
ArtieReus, line 89: Do you really need this?

ArtieReus (changes requested):

ArtieReus (approved):
lgtm

PR #271, opened 2024-07-30
[opentelemetry-operator] add annotation for helm, bump chart

timojohlo:
## Pull Request Details

Adding annotations for CRDs which has to be done manually according to the [documentation](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/UPGRADING.md#0560-to-0570).

PR #282, opened 2024-08-07
template namespace, enable creating crds under templates

timojohlo:
## Pull Request Details

- Removing hard-wired namespace from templates
- removing labels and helm annotations as they will be added automatically 
- enable creating crds under `/templates` instead of custom crds


opentelemetry-operator/chart/crds/crd-opentelemetrycollector.yaml
```diff
@@ -6,10 +6,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.14.0
-    meta.helm.sh/release-name: opentelemetry-operator
-    meta.helm.sh/release-namespace: otel
-  labels:
-    app.kubernetes.io/managed-by: Helm
```
viennaa, line 12: as of #271 this was needed, no?

opentelemetry-operator/chart/values.yaml
```diff
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 opentelemetry-operator:
+  crds:
+    create: false
```
viennaa, line 6: This is disabled, because we're deploying the CRDs dedicatedly?

Why don't we use the operator functionality here and drop our manual adding of CRDs?

viennaa (commented):
Thanks for providing this PR.

You don't need to include the `bump chart` to every change, as every change should have a bumped chart by default.

2 questions inline.

opentelemetry-operator/chart/values.yaml
timojohlo: Yes, that is the reason.

I understood that it is [best practice](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) to store them in a /crds folder rather than in the /template folder

opentelemetry-operator/chart/templates/smon.yaml
```diff
@@ -21,7 +21,7 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/component: opentelemetry-collector
-      app.kubernetes.io/instance: otel.metrics
+      app.kubernetes.io/instance: {{ .Release.Namespace }}.metrics
```
viennaa, line 24: should we be a little bit more explicit here? 

it can be any arbitrary namespace, not sure if I would connect `.metrics` to otel.

Just an idea:

```suggestion
      app.kubernetes.io/instance: {{ .Release.Namespace }}.opentelemetry.metrics
```

viennaa (commented):
Thanks for switching the crd creation. We got to see how this behaves in the future with CRD upgrades, but for now it is good to go :).

Little naming nit to be checked, then we're good to go

timojohlo:
> Thanks for switching the crd creation. We got to see how this behaves in the future with CRD upgrades, but for now it is good to go :).
> 
> Little naming nit to be checked, then we're good to go

Good point. Granted and added for `filelog-smon.yaml` as well.

viennaa (approved):
Cool!

PR #284, opened 2024-08-08
feat(kube-monitoring): create plutono datasource secret

richardtief:
# Submit a pull request

is true:

5. All tests pass.

---

Replace any :question: below with information about your pull request.

## Pull Request Details

This pull request adds an optional secret to the kube-monitoring chart that contains the data source configuration for Plutono to import.

## Breaking Changes

None

## Issues Fixed

None

## Other Relevant Information

:question:

### DIFF
--- a/kube-monitoring/charts/templates/_datasource.tpl
+++ b/kube-monitoring/charts/templates/_datasource.tpl
@@ -1,0 +1,70 @@
+# Configuration file version
+apiVersion: 1
+
+# List of data sources to delete from the database.
+deleteDatasources:
+#  - name: data-source-name
+#    orgId: 1
+
+# List of data sources to insert/update depending on what's
+# available in the database.
+datasources:
+  # <string, required> Sets the name you use to refer to
+  # the data source in panels and queries.
+  - name: {{ printf "%s-%s" $.Release.Name "prometheus" }}
+    # <string, required> Sets the data source type.
+    type: prometheus
+    # <string, required> Sets the access mode, either
+    # proxy or direct (Server or Browser in the UI).
+    # Some data sources are incompatible with any setting
+    # but proxy (Server).
+    access: proxy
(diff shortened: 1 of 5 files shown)


kube-monitoring/charts/templates/_datasource.tpl
```diff
@@ -0,0 +1,70 @@
+# Configuration file version
+apiVersion: 1
+
+# List of data sources to delete from the database.
+deleteDatasources:
```
viennaa, line 5: is this necessary if the list is empty?

kube-monitoring/charts/templates/_datasource.tpl
```diff
@@ -0,0 +1,70 @@
...
+    # <string> Sets the database user, if necessary.
+    user:
+    # <string> Sets the database name, if necessary.
+    database:
+    # <bool> Enables basic authorization.
+    basicAuth:
+    # <string> Sets the basic authorization username.
+    basicAuthUser:
+    # <bool> Enables credential headers.
+    withCredentials:
```
viennaa, line 40: since they are not templated they always will be empty. do we need them or can we comment them out?

viennaa (commented):
Thanks for enabling the datasource. Some questions to empty values inline.

kube-monitoring/charts/templates/_datasource.tpl
richardtief: I intend to leave each field here, even if it is not used, to facilitate additional configurations

kube-monitoring/charts/templates/_datasource.tpl
richardtief: You're right, this could be a template. But this is the first iteration and I would leave it open as a feature moving forward.

viennaa (approved):

PR #308, opened 2024-08-14
feat(heureka/ui): Panelmanager a servicedetails

TilmanHaupt:
## Pull Request Details

Added service detail panel and a panel manger and classes which ensure a greenhouse expirence for heureka. It is ensured that only one panel is shown at one time.

## Issues Fixed

closes https://github.com/cloudoperators/heureka/issues/124

### DIFF
--- a/heureka/ui/src/lib/store.js
+++ b/heureka/ui/src/lib/store.js
@@ -6,6 +6,7 @@
 import { createStore } from "zustand"
 import { devtools } from "zustand/middleware"
 import { produce } from "immer"
+import constants from "../components/shared/constants"
 
 export default (options) =>
   createStore(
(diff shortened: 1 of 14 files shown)


heureka/ui/src/lib/store.js
```diff
@@ -14,6 +14,10 @@ export default (options) =>
       queryClientFnReady: false,
       endpoint: options?.apiEndpoint,
 
+      showPanel: null,
```
hodanoori, line 17: You can use the constant variable here.

hodanoori (changes requested):

hodanoori (approved):
lgtm

Issue #244, opened 2024-07-25
[opentelemetry] make opentelemetry plugin production-ready


timojohlo:
## Context 
The initial version of the OpenTelemetry Plugin is deployed. In order for users to deploy their own version we want to work towards a production-readyness. 

## Acceptance Criteria 
- [x] add documentation on how to configure the OTel Plugin
- [x] check for dependencies and remove dependencies where possible (e.g. CertManager)
- [x] Rename labels in PluginConfiguration and add description for accessibility
- [x] add helm-chart testing

PR #320, opened 2024-08-15
[opentelemetry] add plugin config to enable or disable metri…

timojohlo:
## Pull Request Details

This PR introduces the following changes
- allow to disable logs collecting in the PluginDefinition (default: enabled)
- allow to disable metrics collecting in the PluginDefinition (default: enabled)
- rename the Plugin to OpenTelemetry (rm the postfix -operator), add description in UI
- Add parameters to the Plugin Documentation

## References
- #194 
- #244


opentelemetry/README.md
```diff
@@ -0,0 +1,50 @@
...
+---
+
+This Plugin is intended for ingesting, generating, collecting, and exporting telemetry data (metrics, logs, and traces). 
+
+Components included in this Plugin:
+
+- [Collector](https://github.com/open-telemetry/opentelemetry-collector)
+- [Filelog Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver)
+- [OpenSearch Exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/opensearchexporter)
+- ServiceMonitor to scrape logs statically into prometheus infra-frontend
```
richardtief, line 12: Is the ServiceMonitor a core component of the plugin? If yes or you want to mention it, the Prometheus part should be configurable. Are those just performance metrics or really the logs processed?
```suggestion
- ServiceMonitor to ingest the metrics of the OpenTelemetry Collector in Prometheus 
```

opentelemetry/chart/templates/logs-collector.yaml
```diff
@@ -3,6 +3,7 @@ SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse c
 SPDX-License-Identifier: Apache-2.0
 */}}
 
+{{- if index .Values "open_telemetry" "LogsCollector" "enabled" }}
```
richardtief, line 6: I'm not sure why you are using the `index` function here. To align with [best practices](https://helm.sh/docs/chart_best_practices/values/#naming-conventions), this should be camelCase. Like so
```suggestion
{{- if .Values.openTelemetry.LogsCollector.enabled }}
```

The renaming should then match everywhere.

opentelemetry/chart/templates/metrics-collector.yaml
```diff
@@ -3,6 +3,7 @@ SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse c
 SPDX-License-Identifier: Apache-2.0
 */}}
 
+{{- if index .Values "open_telemetry" "MetricsCollector" "enabled" }}
```
richardtief, line 6: Same here. The index function is not needed. 
```suggestion
{{- if .Values.openTelemetry.MetricsCollector.enabled }}
```

richardtief (changes requested):
Thanks, I like it. Some comments below.

opentelemetry/plugindefinition.yaml
```diff
@@ -4,16 +4,27 @@
...
+    name: opentelemetry
 spec:
-    version: 0.2.6
-    description: opentelemetry-operator
+    version: 0.2.7
+    displayName: OpenTelemetry
+    description: Observability framework for instrumenting, generating, collecting, and exporting telemetry data such as traces, metrics, and logs.
     icon: https://raw.githubusercontent.com/cloudoperators/greenhouse-extensions/main/opentelemetry-operator/logo.png
     helmChart:
         name: opentelemetry-operator
```
viennaa, line 14: Doesn't this need to change too?

```suggestion
        name: opentelemetry
```

opentelemetry/plugindefinition.yaml
```diff
@@ -4,16 +4,27 @@
...
         repository: oci://ghcr.io/cloudoperators/greenhouse-extensions/charts
-        version: 0.2.6
+        version: 0.2.7
     options:
+        - default: true
+          description: Enable Logs Collector with default settings
+          name: open_telemetry.LogsCollector.enabled
```
viennaa, line 20: See Richards comment below, all CamelCase please. Also below.

```suggestion
          name: openTelemetry.LogsCollector.enabled
```

viennaa (changes requested):
Thanks for adding the README, appreciated :)

Comments inline.

opentelemetry/plugindefinition.yaml
richardtief: If yes, the please change the chart, too https://github.com/cloudoperators/greenhouse-extensions/blob/a751d89040a09b11ef5c66a4a71506bd8c085d3f/opentelemetry/chart/Chart.yaml#L6

.. and all the gh actions mentioning it.

opentelemetry/README.md
```diff
@@ -0,0 +1,50 @@
...
+`open_telemetry.prometheus`               | Label for Prometheus Service Monitoring | string | `false` | 
+`open_telemetry.podMonitor.enabled`       | Check to enable the Pod Monitor | bool | `false` | 
+`opentelemetry-operator.admissionWbhooks.certManager.enabled` | Check to use certManager for generating self-signed certificates | bool | `false` | 
+`opentelemetry-operator.admissionWebhooks.autoGenerateCert.enabled` | Check to use Helm to create self-signed certificates | bool | `false` | 
+`opentelemetry-operator.admissionWebhooks.autoGenerateCert.recreate` | Recreate the cert after a defined period (certPeriodDays default is 365) | bool | `false` | 
+`opentelemetry-operator.kubeRBACProxy.enabled` | Check to enable kube-rbac-proxy for OpenTelemetry | bool | `false` | 
+`opentelemetry-operator.manager.prometheusRule.defaultRules.enabled` | Check to enable default rules for monitoring the manager | bool | `false` | 
+`opentelemetry-operator.manager.prometheusRule.enabled` | Check to enable rules for monitoring the manager | bool | `false` | 
+`opentelemetry-operator.manager.serviceMonitor.enabled` | Check to enable the Service Monitor | bool | `false` |
```
viennaa, line 46: Applies to the whole list, all `Check` items. I'd rather use a format like

```Activates Prometheus Monitoring of Collector```

Check and Enable/Disable would be actually obvious. So rather try to explain the functionality with different words. Going with the PodMonitor example it is not self explanatory to state: ```PodMonitor.enabled: Enables PodMonitor.```
General: use different wording

viennaa (changes requested):
Some comments on the README content

opentelemetry/README.md
timojohlo: You are right. I will mention it here and the detailed configuration has to be implemented in the upcoming sprints #325

opentelemetry/plugindefinition.yaml
viennaa: we can drop this, it is to explain which upstream chart is used, so good from my side.

viennaa (approved):
Looks good, thanks.

IvoGoman (approved):

richardtief (approved):

PR #333, opened 2024-08-21
feat(heureka): Remove expensive live search  & improve styling

hodanoori:
## Pull Request Details

- As currently live search handling is very expensive with too many API calls, I deactivated that, to be implementated later with having a proper store state update in place
- Give the min column to columns which contains count number in services and components view

### DIFF
--- a/heureka/ui/src/components/filters/FilterSelect.jsx
+++ b/heureka/ui/src/components/filters/FilterSelect.jsx
@@ -50,13 +50,14 @@ const FilterSelect = ({ entityName, isLoading }) => {
     handleFilterAdd(value)
   }
 
-  const handleSearchChange = (value) => {
+  // TODO: The live search should be implemented after having store update mechanism in place
+  /*const handleSearchChange = (value) => {
     // Debounce search term to avoid unnecessary re-renders
     const debouncedSearchTerm = setTimeout(() => {
       setSearchTerm(entityName, value.target.value)
-    }, 2000) //Every 2 seconds to avoid unnecessary API calls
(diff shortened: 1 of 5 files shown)


heureka/ui/src/components/filters/FilterSelect.jsx
```diff
@@ -54,7 +54,7 @@ const FilterSelect = ({ entityName, isLoading }) => {
     // Debounce search term to avoid unnecessary re-renders
     const debouncedSearchTerm = setTimeout(() => {
       setSearchTerm(entityName, value.target.value)
-    }, 500)
+    }, 2000) //Every 2 seconds to avoid unnecessary API calls
```
edda, line 57: Do you really want to make the user wait 2 seconds before the search is even started? That will not feel good. If you want to avoid unnecessary API calls then I would remove the `onChange` handler and search only on "return" or search button press.

edda (changes requested):

heureka/ui/src/components/filters/FilterSelect.jsx
hodanoori: Since currentlz manz requests are sent while searching for a specific CVE I increased this time. But I wouldnät delete it, since CVEs are often long, this allows users to see a filtered list even with just a part of the CVE. What do you think? Should I still remove onChange func?

hodanoori: I deactivate it, just handle search on Enter and behind search button for now.

edda (approved):

PR #345, opened 2024-08-23
Feat/bump exposed service

uwe-mayer:
# Submit a pull request

This bumps the version of exposed-services with the new reference to `owner-info` Chart


uwe-mayer:
wrongly branched

PR #352, opened 2024-08-26
chore(alerts-ui): Remove Hard-Integrated Authentication

andypf:
### Summary

This PR refactors the Alerts UI (Supernova) by removing the hard-integrated authentication. Supernova doesn't actually require full authentication functionality. Instead of relying on authentication, Supernova should rely on a property called `predefinedFilters` or an event such as `"SUPERNOVA_UPDATE_PREDEFINED_FILTERS"` to manage filter options. 

For now, we've removed the integration of the authentication app, but Supernova still listens to the `"AUTH_UPDATE_DATA"` event to get the user's teams and groups for preset filters.

### Key Benefits

- **Eliminates Unnecessary Dependencies**: By removing the direct integration with the authentication app, the dependency on the authentication app becomes obsolete. This is particularly beneficial since the authentication app is not published to npm.
  
- **Standalone Operation**: This refactor is a step toward enabling Supernova to run standalone without any dependencies. It simplifies the future development of a Dockerfile that can run the app independently.


alerts/ui/src/AppContent.jsx
```diff
@@ -48,17 +43,9 @@ const AppContent = () => {
...
-    })
-  }, [authError])
-
   useEffect(() => {
     // since the API call is done in a web worker and not logging aware, we need to show the error just in case the user is logged in
-    if (!alertsError || !loggedIn) return
+    if (!alertsError) return
 
     // if user uses firefox warn to activate `allow_client_cert`. Should be enough to do it just here since the API call is done in a web worker and nothing else will be loaded until the alerts are loaded
     const isFirefox = navigator.userAgent.toLowerCase().includes("firefox")
```
edda, line 51: I just realized that this is also very specific to our auth setup. We should move this out of Supernova as well.

alerts/ui/src/components/CustomAppShell.jsx
```diff
@@ -62,10 +47,10 @@ const CustomAppShell = ({ children }) => {
 
   return (
     <AppShell
-      pageHeader={pageHeader}
+      pageHeader={<PageHeader heading="Supernova" />}
       embedded={embedded}
-      topNavigation={topNavigation}
     >
+      {topNavigation}
```
edda, line 53: The way the `topNavigation` was passed as a prop to the Appshell was correct. If this didn't work for you then we need to upgrade `juno-ui-components` to the latest version

alerts/ui/docker/Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+COPY --from=build /app /usr/share/nginx/html
+
+ENV ENDPOINT=""
+ENV THEME="theme-dark"
+ENV ENDPOINT="" 
+ENV FILTER_LABELS="" 
+ENV SILENCE_EXCLUDED_LABELS=""
+ENV SILENCE_TEMPLATES=""
+
+# Create Entrypoint which creates /app/appProps.js from ENVs
```
edda, line 66: Can we provide a sample appProps.js that contains dummy examples of all the above settings formatted correctly as expected? We state in the Readme what we expect but I think it would really help if there'd be a sample config that people can take and edit.

alerts/ui/docker/README.md
```diff
@@ -0,0 +1,37 @@
...
+Ensure you are in the `alerts/ui` directory, then build the Docker image with:
+
+```bash
+docker build -t supernova -f docker/Dockerfile .
+```
+
+## Run the Container
+
+```bash
+docker run -e ENDPOINT="https://alertmanager_url/api/v2" -e THEME="theme-light" -p 3010:80 supernova
```
edda, line 16: Remove passing of theme env from the default run command. We can add a second example that shows how env variables can be passed:
```suggestion
docker run -e ENDPOINT="https://alertmanager_url/api/v2" -p 3010:80 supernova
```

edda: Can the env variables also be passed from a file? The more complex ones like silence templates will be hard to do in a CLI command

edda (changes requested):

edda (approved):

ArtieReus (approved):
lgtm

alerts/ui/docker/Dockerfile
```diff
@@ -0,0 +1,132 @@
+# SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse contributors
+# SPDX-License-Identifier: Apache-2.0
+
+FROM node:20-alpine as build
+
+# we need jq to read and manipulate package.json
+RUN apk add --no-cache jq
```
kengou, line 7: I would like to see some image update during build time.
```suggestion
RUN apk add --no-cache jq \
  && apk upgrade --no-cache --no-progress \
  && apk del --no-cache --no-progress apk-tools alpine-keys
```

alerts/ui/docker/Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+  "silenceTemplates": ""
+}
+EOF
+
+# clean up
+RUN rm /app/entry_file
+
+FROM nginx:alpine
+
+RUN apk add --no-cache jq
```
kengou, line 72:
```suggestion
RUN apk add --no-cache jq \
  && apk upgrade --no-cache --no-progress \
  && apk del --no-cache --no-progress apk-tools alpine-keys
```

IvoGoman (approved):

.github/workflows/build-push-supernova-image.yaml
```diff
@@ -0,0 +1,108 @@
+# Run it locally with act
+#  1. Install act:
+#     `brew install act`
+#  2. Create a .secret file with the following content:
+#     `GITHUB_TOKEN=your_github_token`
+#  WORKFLOW_DISPATCH
+#     `act workflow_dispatch  --container-architecture linux/amd64 -P default=catthehacker/ubuntu:act-latest -W .github/workflows/build-push-supernova-image.yaml`
+
+name: Supernova UI Image
```
kengou, line 9:
```suggestion
name: Build Supernova UI Image
```

.github/workflows/build-push-supernova-image.yaml
```diff
@@ -0,0 +1,108 @@
...
+      # with sigstore/fulcio when running outside of PRs.
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # This action enables you to SIGN and VERIFY container images using cosign
+      # https://github.com/sigstore/cosign-installer
+      - name: Install cosign
```
kengou, line 42: Why `cosign` is installed when not using it and signing the image like here: 
https://github.com/cloudoperators/greenhouse/blob/cd0e226dc8e272106d15f05b60d37667a2e7dc0a/.github/workflows/docker-build.yaml#L128-L136

.github/workflows/build-push-supernova-image.yaml
```diff
@@ -0,0 +1,108 @@
...
+        with:
+          context: ${{ env.PACKAGE_PATH }}
+          file: ${{ env.PACKAGE_PATH }}/docker/Dockerfile
+          # change to true to push
+          push: false
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          platforms: |
+            linux/amd64
+            linux/arm64
```
kengou, line 108: I'm missing the privy check from the gh-action like this:
https://github.com/cloudoperators/greenhouse/blob/cd0e226dc8e272106d15f05b60d37667a2e7dc0a/.github/workflows/docker-build.yaml#L138-L180

kengou (changes requested):

.github/workflows/build-push-supernova-image.yaml
ArtieReus: make sence :)

.github/workflows/build-push-supernova-image.yaml
ArtieReus: This is planed for the second iteration. The build and push step is for now deactivated. We want just see the action in action :)

.github/workflows/build-push-supernova-image.yaml
ArtieReus: This is planed for the second iteration. The build and push step is for now deactivated. We want just see the action in action :)

PR #372, opened 2024-09-04
feat(kube-monitoring): adding default pmon

viennaa:
services with a port metrics are getting collected now

### DIFF
--- a/kube-monitoring/charts/templates/pmon.yaml
+++ b/kube-monitoring/charts/templates/pmon.yaml
@@ -1,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: {{ $.Release.Name }}-sd
+  labels:
+    plugin: {{ $.Release.Name }} 
+spec:
+  namespaceSelector:
+    any: true
+  podMetricsEndpoints:
+  - port: metrics
+  selector:
+    matchExpressions:
+    - key: foo
+      operator: DoesNotExist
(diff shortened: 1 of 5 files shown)


kube-monitoring/charts/templates/pmon.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: {{ $.Release.Name }}-sd
```
ibakshay, line 4: Does "sd" stand for "service discovery" here?

kube-monitoring/charts/templates/tests/test-kube-monitoring-config.yaml
```diff
@@ -53,6 +53,9 @@ data:
     @test "Verify creation of the prometheus-{{ .Release.Name }} statefulset" {
         verify "there is 1 statefulset named 'prometheus-{{ .Release.Name }}'"
     }
+    @test "Verify creation of the {{ .Release.Name }}-sd Podmonitor" {
```
ibakshay, line 56: Adding a blank line to be consistent like other test cases. 
```suggestion

    @test "Verify creation of the {{ .Release.Name }}-sd Podmonitor" {
```

kube-monitoring/charts/templates/pmon.yaml
viennaa: yes

richardtief:
Where does this feature request come from? This will lead to multiple pod scrapes if we install another Prometheus, for example.

PR #373, opened 2024-09-04
feat(heureka): Edit service owners

hodanoori:
## Pull Request Details

- **Note**: The API supports adding or removing only one user at a time from the owner list.
- On the service details page, to add a new owner, click the "+" icon in the "Owner" cell, select a user from the combo box, and click "Add".  The newly added owner will then appear as a pill.
- Existing owners are displayed as pills with a remove (x) option. When attempting to remove an owner, a confirmation dialog is displayed to ensure the user wishes to proceed.
- If a user switches to another service's details while in edit mode to add an owner, the edit mode is exited, and the newly selected service details are shown in display mode.
- A success message is displayed upon successfully adding or removing an owner.


TilmanHaupt (commented):
If you change between the details views the pills stay from the old detail. I would think about cleaning them. After changing the detailView. Also fixing size of the header-colum could be considered as talked in the sync.

heureka/ui/src/components/services/ServicesDetails.jsx
```diff
@@ -3,127 +3,274 @@
...
+        {confirmRemoveOwner && (
+          <>
+            <Modal
+              cancelButtonLabel="Cancel"
+              confirmButtonLabel="Yes, Proceed"
+              open={!!confirmRemoveOwner}
+              onCancel={() => setConfirmRemoveOwner(null)}
+              onConfirm={() => handleRemoveOwner(confirmRemoveOwner)}
+            >
+              <p>Are you sure you want to remove this owner?</p>
```
edda, line 219: It's always better to repeat to the user what exactly is getting removed. So pass the user name to the modal and then change the text to `Are you sure you want to remove { ownerName } from owners?`

edda (changes requested):

heureka/ui/src/components/services/ServicesDetails.jsx
```diff
@@ -3,127 +3,274 @@
...
+                <Icon
+                  color="jn-global-text"
+                  icon="addCircle"
+                  onClick={() => setShowComboBox(true)}
+                  className="ml-2"
+                />
+              </Stack>
+            )}
+
+            {showComboBox && (
```
edda, line 174: When the add user combobox is shown and the user clicks on another service in the services list then the details view is updated but the combobox is still shown. This is confusing. 
Either prevent list clicks when the combobox is shown or ensure that the "edit mode", i.e. combobox is visible is reset when the user clicks on another service in the list.

edda (changes requested):

edda (approved):

PR #381, opened 2024-09-06
fix(logshipper): default output did not match

IvoGoman:
The introduction of the `default_` prefix broke the log routing to the output.


ibakshay (approved):
Good find! 🚀

PR #394, opened 2024-09-17
Otel testing

Kuckkuck:
- ceph parsing
- new image registry
 A lot of commits for testing


opentelemetry/chart/templates/smon-filelog.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
   annotations:
   name: opentelemetry-collector-logs
   labels:
-    prometheus: "{{ .Values.openTelemetry.prometheus }}"
+    prometheus: infra-frontend
```
viennaa, line 12: I would rather have this specced in the values, this is very static and will involve touching soon.

opentelemetry/chart/values.yaml
```diff
@@ -1,17 +1,33 @@
...
+    create: true
+    certManager:
+      enabled: false
+    autoGenerateCert:
+      enabled: true
+      recreate: false
   manager:
     collectorImage:
-      repository: ghcr.io/cloudoperators/opentelemetry-collector-contrib
+      repository: keppel.eu-de-1.cloud.sap/ccloud/opentelemetry/collector-image
```
viennaa, line 26: make this configurable. it should default to cloudoperators, we add this as a value in secrets for ccloud deployment.

opentelemetry/chart/values.yaml
```diff
@@ -1,17 +1,33 @@
...
       #repository: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib
-      tag: main
       #tag: 0.104.0
+      #tag: main
+      tag: latest
```
viennaa, line 30: can we go with versions from Chart.yaml here? same applies to opentelemetry-operator

viennaa (changes requested):
Thanks for adding this PR. Some comments are inline.

Please squash the commits to at most 5 and give it a meaningful message like

* dashboard added
* crds added
* image location changed

opentelemetry/chart/values.yaml
Kuckkuck: switched to docker image as default

viennaa (approved):
Review took place via phone. Good to go.

PR #398, opened 2024-09-17
Otel updates for scraping

Kuckkuck:
- bump version
- different label for prometheus scraping
- fixed version for otel-contrib
- 2 years for certs
- enable podMonitor

### DIFF
--- a/opentelemetry/chart/values.yaml
+++ b/opentelemetry/chart/values.yaml
@@ -18,9 +18,7 @@ opentelemetry-operator:
     create: true
     certManager:
       enabled: true
-      duration: "17280h"
     autoGenerateCert:
-      enabled: false
       recreate: false
   manager:
     collectorImage:
@@ -47,10 +45,6 @@ openTelemetry:
   logsCollector:
     enabled: true
   prometheus:
-  podMonitor:
-    enabled: true
-  serviceMonitor:
-    enabled: false
 testFramework:
   enabled: true
   image:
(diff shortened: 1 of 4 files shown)


opentelemetry/chart/values.yaml
```diff
@@ -44,10 +45,10 @@ openTelemetry:
   cluster:
   region:
   logsCollector:
-    enabled: false
+    enabled: true
   prometheus:
   podMonitor:
-    enabled: false
```
IvoGoman, line 50: The `logsCollector` and `podMonitor` are also options in the `PluginDefinition` where they are defaulted to true, false respectively.  These values in the `values.yaml` have no effect in this case and the values should be configured on the Plugin instead.

opentelemetry/chart/values.yaml
```diff
@@ -17,14 +17,15 @@ opentelemetry-operator:
     failurePolicy: 'Ignore'
     create: true
     certManager:
-      enabled: false
-    autoGenerateCert:
       enabled: true
+      duration: "17280h"
+    autoGenerateCert:
+      enabled: false
       recreate: false
```
IvoGoman, line 24: Same as for the comment below. These values are part of the PluginDefinition's `options` and should be configured directly on the Plugin.

opentelemetry/chart/values.yaml
Kuckkuck: I have removed these defaults.

opentelemetry/chart/values.yaml
Kuckkuck: I have removed these defaults.

IvoGoman (approved):

PR #399, opened 2024-09-18
plutono: serve from api server sub path

richardtief:
To make it functional with the service proxy. Fixes https://github.com/cloudoperators/greenhouse/issues/407


ibakshay (approved):
great find 🚀

PR #428, opened 2024-10-04
[opentelemetry] test values, bump chart

Kuckkuck:
Adding ci folder for testing
moving a lot of defaults to test-values.yaml
adding missing defaults.
labels for prometheus


ibakshay:
@Kuckkuck, I cannot see the ` test-values.yaml` in the ci folder. Did you forget to push this change?

opentelemetry/chart/templates/logs-collector.yaml
```diff
@@ -39,6 +39,7 @@ spec:
       port: 9999
 {{- end }}
   image: {{ index .Values "opentelemetry-operator" "manager" "collectorImage" "repository" }}:{{ index .Values "opentelemetry-operator" "manager" "collectorImage" "tag" }}
+  upgradeStrategy: automatic
```
viennaa, line 42: is this really needed? it reads as if this is the default

https://pkg.go.dev/github.com/open-telemetry/opentelemetry-operator#section-readme

> The default and only other acceptable value for .Spec.UpgradeStrategy is automatic.

opentelemetry/chart/templates/smon.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
   annotations:
   name: opentelemetry-collector-metrics
   labels:
+    plugin: "{{ .Values.openTelemetry.prometheus }}"
```
viennaa, line 12: How does openTelemetry.prometheus map to the otel plugin name?

viennaa (commented):
Thanks for updating the chart. Please check my remarks.

opentelemetry/chart/templates/smon.yaml
Kuckkuck: at the moment it has to be set in the plugin page.

ibakshay (commented):
The helm test is passing now! 🚀

viennaa (changes requested):
👍 for including and using the label helper.

can you integrate the functionality to add custom labels too, please?

example:
https://github.com/cloudoperators/greenhouse-extensions/blob/main/thanos/charts/templates/store-svc.yaml#L14-L16

viennaa (approved):
Thanks! Go.

PR #459, opened 2024-10-22
[opentelemetry] add custom labeling for prometheus

timojohlo:
## Pull Request Details
This serves as a baseline for Prometheus labeling. The labels will be injected through secrets and the PluginPresets


opentelemetry/chart/templates/_helpers.tpl
```diff
@@ -7,8 +7,12 @@ Generic plugin name
 
 {{/* Generate plugin specific labels */}}
 {{- define "plugin.labels" -}}
-plugindefinition: opentelemetry 
-plugin: {{ $.Release.Name }}
+plugindefinition: opentelemetry
+{{- if $.Values.prometheusRules.ruleSelectors }}
+{{- range $i, $target := $.Values.prometheusRules.ruleSelectors }}
+{{ $target.name | required (printf "$.Values.prometheusRules.ruleSelector.[%v].name missing" $i) }}: {{ tpl ($target.value | required (printf "$.Values.prometheusRules.ruleSelector.[%v].value missing" $i)) $ }}
```
richardtief, line 13: We can keep it simple. Like so:
```suggestion
{{- with $.Values.openTelemetry.prometheus.additionalLabels }}
  {{- tpl (toYaml . | nindent 4) $ }}
{{- end }}
```

richardtief (commented):
Thank you! Got some suggestions below.

opentelemetry/chart/templates/_helpers.tpl
```diff
@@ -7,8 +7,10 @@ Generic plugin name
 
 {{/* Generate plugin specific labels */}}
 {{- define "plugin.labels" -}}
-plugindefinition: opentelemetry 
-plugin: {{ $.Release.Name }}
+plugindefinition: opentelemetry
+{{- with $.Values.openTelemetry.prometheus.additionalLabels }}
+{{ tpl (toYaml . ) $ }}
+{{- end }}
```
richardtief, line 13: Just realized that all resources get the additional label. But that only needs to be added to the `ServiceMonitor`, `PodMonitor`, `PrometheusRules`, etc. - e.g. [here](https://github.com/cloudoperators/greenhouse-extensions/blob/638099cc7853b3940d4e7443a5931a3fcddbe240/opentelemetry/chart/templates/smon.yaml#L7-L16)
```yaml
kind: ServiceMonitor
metadata:
  annotations:
  name: opentelemetry-collector-metrics
  labels:
    {{- include "plugin.labels" . | nindent 4 }}
    {{- with .Values.openTelemetry.prometheus.additionalLabels }}
    {{ tpl (toYaml . ) $ }}
    {{- end }}
    {{- if .Values.customLabels }}
    {{ toYaml .Values.customLabels | nindent 4 }}
    {{- end }}
spec:
```

```suggestion
```

richardtief (changes requested):
Thanks. Some more hints below.

PR #460, opened 2024-10-23
chore(opentelemetry): revise prometheus monitoring 

richardtief:
- grouped the prometheus-related values into one directive
- PodMonitor and ServiceMonitor are enabled by default
- Added indication whether `monitoring.coreos.com/v1` is available on the cluster
- plugindefinition updated


timojohlo (approved):
Thank you for taking this!

PR #478, opened 2024-11-04
[opentelemetry] adding logs label, fixing ceph parsing

Kuckkuck:
- Fixed ceph log line parsing for resource.attributes["app.label.component"] == "cephobjectstores.ceph.rook.io"
- added label for log.type
- increase memory limit
- more Kubernetes labels from the Kubernetes metadata service


opentelemetry/chart/templates/logs-collector.yaml
```diff
@@ -56,13 +56,19 @@ spec:
     receivers:
       journald:
         directory: /var/log/journal
+        operators:
+          - id: journal-label
+            type: add
+            field: attributes["log.type"]
+            value: "journald"
```
viennaa, line 63: are the quotes necessary?

opentelemetry/chart/templates/logs-collector.yaml
```diff
@@ -128,6 +134,9 @@ spec:
           - action: insert
             key: region
             value: "${region}"
+          - action: insert
+            key: log.type
+            value: "k8sevents"
```
viennaa, line 139: same question :)

viennaa (commented):
Small nit from my side.

opentelemetry/chart/templates/logs-collector.yaml
Kuckkuck: same answer ;-)

Issue #356, opened 2024-08-28
Improving the documentation of the observability package 


richardtief:
## Context

For observability-related releases/plugins, there should be an accessible description that clearly describes the scope and value proposition of each release. For this purpose, business language should be used that emphasises less the technical specifications. All this in conjunction with a Quick Start Guide to use it standalone and/or in conjunction with Greenhouse. 

## Objectives

Why is it necessary?
What is it for? 
What does it do? 
How can I use it? 
What are the dependencies?
What is it not?
How can it be expanded?
Where can I contribute?

richardtief:
Done. Foundation set to iterate over.

PR #479, opened 2024-11-04
docs(opentelemetry): add more context

richardtief:
Part of #356

### DIFF
--- a/opentelemetry/README.md
+++ b/opentelemetry/README.md
@@ -2,14 +2,50 @@
 title: OpenTelemetry
 ---
 
-This Plugin is intended for ingesting, generating, collecting, and exporting telemetry data (metrics, logs, and traces). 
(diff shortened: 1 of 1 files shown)


opentelemetry/README.md
```diff
@@ -2,14 +2,50 @@
...
+
+... 
+
+**Step 2:**
+
+...
+
+**Step 3:**
+
+...
```
richardtief, line 46: @timojohlo @Kuckkuck Could you please fill in the blanks here? I should describe at a high level how to easily get started with the plugin you developed here. Thanks

richardtief: If you need inspiration: https://github.com/cloudoperators/greenhouse-extensions/blob/polish-plugin-docs/kube-monitoring/README.md#quick-start

timojohlo (approved):
Thanks for driving this!

PR #490, opened 2024-11-12
chore(deps): run helm dep up for kube-monitoring and alerts

richardtief:


IvoGoman:
Can you update the PluginDefinitions with the new chart versions?

IvoGoman (approved):

PR #507, opened 2024-11-14
docs(opentelemetry) add architecture img

timojohlo:
## Pull Request Details

add OpenTelemetry architecture image


timojohlo:
> Can you check the resolution of the image? It seems rather low, when looking at it in a dedicated tab. Also please select a white background, as the kubernetes logo is in front of a transparent background.
> 
Agreed, I will increase the resolution and make the background transparent

> I would split the picture a little more. Like so: Have the collector part with the processors only and the receivers and exporters are left and right of the collector box, no?
> 
> Technically these are external components which are pushing to the collector with OTLP. Then the Processor/Pipelining part starts. Then they are shipped to other external components (OpenSearch, Prometheus).

I tend to differ here, as, even though the components are pushing to OTEL, the receivers are specified for each external component and their configuration reside in the collector itself. Meaning: The sources and sinks for data are external, but from which source and to which sink the data is processed and how it is received and exported is configured in the Collector.

viennaa:
Resolution is huuuuge now, like it :D 

> I tend to differ here, as, even though the components are pushing to OTEL, the receivers are specified for each external component and their configuration reside in the collector itself. Meaning: The sources and sinks for data are external, but from which source and to which sink the data is processed and how it is received and exported is configured in the Collector.

Fair point in regards to the configuration. I still would like to see the outer components (sources, sinks) pushing to the respective receivers. Currently the part _how_ comes everything in is missing. This then would be a 1:1 connection, but nevertheless it makes it clearer. Same then also to the right side. For filelog, k8s events etc. this can also be very generic.

IvoGoman (approved):

PR #512, opened 2024-11-15
chore(alerts): adds alertmanager dashboard

richardtief:


alerts/plugindefinition.yaml
```diff
@@ -62,6 +62,10 @@ spec:
       default: true
       required: false
       type: bool
+    - name: alerts.dashboard.create
+      description: Whether to install the default dashboards for self-monitoring Prometheus Alertmanger. 
+      default: false
```
viennaa, line 67: don't we always want this?
```suggestion
      default: true
```

richardtief: Good point. Let's opt out if necessary

alerts/charts/dashboards/alertmanager.json
```diff
@@ -0,0 +1,4164 @@
...
+        "datasource": "-- Plutono --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "description": "Dashboard showing Prometheus Alertmanager metrics for observing status of the cluster and possible debbuging.",
```
ibakshay, line 15:
```suggestion
  "description": "Dashboard showing Prometheus Alertmanager metrics for observing status of the cluster and possible debugging.",
```

ibakshay (changes requested):

alerts/charts/values.yaml
```diff
@@ -45,6 +45,16 @@ alerts:
...
 
+  ## Create curated dashboards for generic monitoring (e.g. monitoring Alertmanager)
+  ##
+  dashboards:
+    create: true
+
+    ## Label selectors for the Plutono dashboards to be picked up by Plutono.
+    plutonoSelectors:
+      - name: plutono-dashboard
+        value: '"true"'
```
viennaa, line 56: do we need double quoting? If so, why?

richardtief: I made it a `string` because label values do not allow boolean values. And Helm will remove the quotes during template. 

`The edited file had a syntax error: unable to decode "edited-file": json: cannot unmarshal bool into Go struct field ObjectMeta.metadata.labels of type string`

https://github.com/cloudoperators/greenhouse-extensions/blob/9f9c7fe3e6ea084bdbf28c0ae62da0dd3fc81bdd/plutono/charts/values.yaml#L935

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

viennaa (approved):
Thanks!

ibakshay (approved):
Lets ship it! 🚀

PR #518, opened 2024-11-19
[opentelemetry] Clean up of the helm values and structure

olandr:
- Removes redundant values from values.yaml
- Refactored and changed a few values of the values.yaml, added comments, updated references.
- Changed filtering logic for alerts
- Bumps version

### DIFF
--- a/opentelemetry/chart/alerts/collector-alerts.yaml
+++ b/opentelemetry/chart/alerts/collector-alerts.yaml
@@ -1,7 +1,7 @@
 groups:
   - name: collector-alerts
     rules:
-  {{- if not (.Values.prometheusRules.disabled.FilelogRefusedLogs | default false) }}
+  {{- if not (has "FilelogRefusedLogs" .Values.openTelemetry.prometheus.rules.disabled) }}
       - alert: FilelogRefusedLogs
         expr: sum(rate(otelcol_receiver_refused_log_records_total{receiver=~"filelog"}[1m])) > 0
         for: 5m
@@ -14,7 +14,7 @@ groups:
           description: Filelog receiver is increasingly rejecting logs
   {{- end }}
 
-  {{- if not (.Values.prometheusRules.disabled.ReceiverRefusedMetric | default false) }}
+  {{- if not (has "ReceiverRefusedMetric" .Values.openTelemetry.prometheus.rules.disabled) }}
       - alert: ReceiverRefusedMetric
         expr: sum(rate(otelcol_receiver_refused_metric_points_total{}[1m])) > 0
         for: 5m
@@ -27,7 +27,7 @@ groups:
           description: Maybe collector has received non standard metric points or it reached some limits
   {{- end }}
 
-  {{- if not (.Values.prometheusRules.disabled.HighCPUUsage | default false) }}
+  {{- if not (has "HighCPUUsage" .Values.openTelemetry.prometheus.rules.disabled) }}
       - alert: HighCPUUsage
         expr: max(rate(otelcol_process_cpu_seconds{}[1m])*100) > 90
         for: 5m
--- a/opentelemetry/chart/ci/test-values.yaml
+++ b/opentelemetry/chart/ci/test-values.yaml
@@ -45,6 +45,12 @@ openTelemetry:
     additionalLabels:
       key1: value1
       key2: value2
+
+    rules:
+      create: true
+      disabled:
+        - FilelogRefusedLogs
+
 testFramework:
   enabled: false
   image:
--- a/opentelemetry/chart/ci/test-values.yaml
+++ b/opentelemetry/chart/ci/test-values.yaml
@@ -29,6 +29,7 @@ opentelemetry-operator:
       enabled: false
   kubeRBACProxy:
     enabled: false
+
 openTelemetry:
   openSearchLogs:
     endpoint: test
@@ -36,6 +37,10 @@ openTelemetry:
     password: test
   cluster: test
   region: test
+  logsCollector:
+    enabled: true
+  metricsCollector:
+    enabled: false
   prometheus:
     additionalLabels:
       key1: value1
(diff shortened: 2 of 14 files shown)


opentelemetry/chart/alerts/collector-alerts.yaml
```diff
@@ -1,7 +1,7 @@
 groups:
   - name: collector-alerts
     rules:
-  {{- if not (.Values.prometheusRules.disabled.FilelogRefusedLogs | default false) }}
+  {{- if not (.Values.opentelemetry.prometheus.rules.disabled.FilelogRefusedLogs | default false) }}
```
viennaa, line 4: This is pretty hard to read, suggesting to simplify this to

```suggestion
  {{- if not ( has 'FilelogRefusedLogs' .Values.opentelemetry.prometheus.rules.disabled }}
```
and use it as sort of a _blacklist_.

`Values.opentelemetry.prometheus.rules.disabled` needs to be used as a list, instead of dict then.

Applies to all alerts here.

opentelemetry/chart/ci/test-values.yaml
```diff
@@ -29,17 +29,26 @@ opentelemetry-operator:
...
+    enabled: true"
+  metricsCollector:
+    enabled: false
   prometheus:
     additionalLabels:
       key1: value1
       key2: value2
+
+    rules:
+      create: false
```
viennaa, line 50: at least one rule should be disabled here for the sake of testing. - and of course this value needs to be enabled, otherwise the whole alerting logic is not tested.

opentelemetry/chart/templates/_helpers.tpl
```diff
@@ -8,24 +8,24 @@ Generic plugin name
...
+{{- if .Values.commonLabels }}
+{{ tpl (toYaml .Values.commonLabels) . }}
 {{- end }}
 {{- end }}
 
 {{/* Generate prometheus specific labels */}}
 {{- define "plugin.prometheusLabels" }}
-{{- if .Values.openTelemetry.prometheus.additionalLabels }}
-{{- tpl (toYaml .Values.openTelemetry.prometheus.additionalLabels) . }}
+{{- if .Values.opentelemetry.prometheus.additionalLabels }}
```
viennaa, line 18: What's the reason for uncamel-casing?

opentelemetry/plugindefinition.yaml
```diff
@@ -4,6 +4,7 @@
 apiVersion: greenhouse.sap/v1alpha1
 kind: PluginDefinition
 metadata:
+    creationTimestamp: null
```
viennaa, line 7: Can be deleted, no?
```suggestion
```

viennaa (changes requested):
Thanks for taking the broom and starting to tidy up. Very much appreciated.

Some remarks are inline and need to be addressed.

Additionally (a little bit more verbose for easier understanding): Why do we need to add the monitoring CRDs here? Wouldn't they conflict with the CRDs provisioned via prometheus-operator?
Since you are checking for the capabilities, pmons and smons will not be deployed, unless the resources are in the cluster. 
This dependency however needs to be written down in the README. In the end deploying the monitoring resources makes no sense, if there is no prometheus operator picking them up. So unless you are deploying the operator too (which we are doing already in the alerts or kube-monitoring plugin), this is not needed.

Reference 1:
https://github.com/cloudoperators/greenhouse-extensions/blob/main/alerts/plugindefinition.yaml#L45-L49

Reference 2:
https://github.com/cloudoperators/greenhouse-extensions/blob/main/kube-monitoring/charts/values.yaml#L65-L68

opentelemetry/chart/templates/_helpers.tpl
olandr: I chose this for the sake of consistency. Plenty of other references used all lowercase, the otel-operator dependency is also named in all lowercase, the lowercase name is the same for the directory of the plugin + the reference for PRs, issues etc.

olandr:
> Why do we need to add the monitoring CRDs here? Wouldn't they conflict with the CRDs provisioned via prometheus-operator?
> So unless you are deploying the operator too (which we are doing already in the alerts or kube-monitoring plugin), this is not needed.

I see. If it is a pre-condition that the prom-operator exists then ofc the CRD will not be needed. The way that I thought this plugin was progressing was unrelated to other plugins (i.e. that there was not a particular dependency tree between plugins).


I will make the README a bit more clear on the requirements here in that case, and remove the redundant CRDs.

opentelemetry/chart/alerts/collector-alerts.yaml
olandr: Agreed, that is clearer to read.

To make the `.Values....rules` clear what is your take on adding this to the `Values.yaml`

opentelemetry/chart/values.yaml
```diff
@@ -87,6 +59,27 @@ openTelemetry:
...
     # plugin: kube-monitoring
     # prometheus: infra
+
+    ## Default rules for monitoring the opentelemetry components
+    rules:
+      ## Enables PrometheusRule resources to be created
+      create: true
+
+      # PrometheusRules to disable
+      disabled: []
```
viennaa, line 69: just beware, that this might break existing deployments, if they have specified a dict.

viennaa (changes requested):
Bump the minor version please.

https://github.com/cloudoperators/greenhouse-extensions/blob/main/opentelemetry/chart/Chart.yaml#L7

Same for the pluginconfig version.

viennaa:
> Bump the minor version please.
> 

Minor please, not patch :)
https://semver.org/

opentelemetry/chart/templates/_helpers.tpl
olandr: Turns out that I was a bit too eager here. I had missed quite a few related references of this particular name. I will revert the casing and then in the future potentially having a dedicated PR for dealing with all cascading/depending references.

viennaa (changes requested):
Please squash the commits to a reasonable amount for the changes in this PR. They also should be sorted topic wise, to make clear, what was introduced by which commit.

Make it like 2-3 commits i'd say.

viennaa (approved):
Thanks, good to go from my side.

timojohlo:
Nice, first PR done I would say 👍

PR #530, opened 2024-11-22
fix(alertts): Add conditional check for Prometheus Operator in test alerts config

ibakshay:
Introduce a conditional check to ensure the test alerts configuration only verifies the deployment of the Prometheus Operator when it is enabled.


viennaa (approved):
Yep.

PR #534, opened 2024-11-26
[WIP] thanos ruler compoment

trouaux:
# Submit a pull request

is true:

5. All tests pass.

---

Replace any ":question:" below with information about your pull request.

## Pull Request Details

:question:

## Breaking Changes

:question:

## Issues Fixed

1. :question:

## Other Relevant Information

:question:


thanos/charts/templates/ruler.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+spec:
+  image: "{{ .Values.thanos.image.repository }}:{{ .Values.thanos.image.tag | default .Chart.AppVersion }}"
+  paused: {{ $.Values.ruler.paused }}
+  logLevel: {{ default "info" $.Values.ruler.logLevel }}
+  {{- if $.Values.ruler.retention }}
+  retention: {{ $.Values.ruler.retention | quote }}
+  {{- end }}
+  {{- if $.Values.rbac.create }}
+  serviceAccountName: {{ $.Values.serviceAccount.name }}
```
viennaa, line 24: Would go with a default here: 
```suggestion
  serviceAccountName: {{ include "release.name" . }}-ruler
```

viennaa (changes requested):
Thanks for tackling your first issue and opening this PR.

We need to change the deployment to a statefulset. Will continue the PR after this change is in.

Please add testing for the respective statefulset to come up.

Please fix the linting test.

thanos/charts/values.yaml
```diff
@@ -103,10 +104,86 @@ thanos:
...
+      kubeSchedulerAlerting: false
+      kubeSchedulerRecording: false
+      kubeStateMetrics: false
+      network: false
+      node: false
+      nodeExporterAlerting: false
+      nodeExporterRecording: false
+      prometheus: false
+      prometheusOperator: false
+      windows: false
```
viennaa, line 164:
```suggestion
```

We can remove this if create is set to false.

viennaa (commented):
Little improvement :)

thanos/charts/values.yaml
```diff
@@ -36,36 +34,39 @@ thanos:
...
     # adding additional arguments
     additionalArgs: []
 
-    deploymentLabels: {} 
+    deploymentLabels: {}
 
     containerLabels:
 
-    #TODO use
+    # TODO use
```
ibakshay, line 55: do we need this comment?

ibakshay (commented):
LGTM! left a minor comment :)

thanos/charts/Chart.yaml
```diff
@@ -3,18 +3,23 @@
...
 appVersion: v0.35.0
 keywords:
   - thanos
   - storage
   - metrics
+dependencies:
+  - name: kube-prometheus-stack
+    alias: prometheus-operator
+    repository: https://prometheus-community.github.io/helm-charts
+    version: 65.4.0
```
viennaa, line 25: take the newer bumped version here pls and `helm dep up` it

thanos/charts/templates/ruler.yaml
```diff
@@ -0,0 +1,32 @@
...
+    {{- include "plugin.labels" . | nindent 4 }}
+    {{- include "thanos.labels" . | nindent 4 }}
+    {{- if .Values.thanos.ruler.labels }}
+    {{ toYaml .Values.thanos.ruler.labels | nindent 4 }}
+    {{- end }}
+spec:
+  image: "{{ .Values.thanos.image.repository }}:{{ .Values.thanos.image.tag | default .Chart.AppVersion }}"
+  ruleSelector:
+    matchLabels:
+      thanos-ruler: {{ .Values.thanos.ruler.matchLabel | default .Release.Name }} # Label for fetching rules
```
viennaa, line 21:
```suggestion
      thanos-ruler: {{ .Values.thanos.ruler.matchLabel | default .Release.Name }}
```

either the comment goes above or we leave it for good. decide if you still want do have it in for understanding purposes.

thanos/charts/values.yaml
viennaa: we leave it in for now, I remember I wanted to look at it.

thanos/charts/templates/ruler.yaml
```diff
@@ -0,0 +1,32 @@
...
+  queryEndpoints:
+    - {{ include "release.name" . }}-query:10902
+  alertmanagersConfig:
+    key: alertManagerConfig.yaml # Key name for alert manager config yaml
+    name: {{ include "release.name" . }}-alertmanager-config # Secret name containing alert manager config
```
viennaa, line 26: do we have this config around?
I think this is something which needs to be added to communicate with the alertmanager. as this is a configmap it should be part of this PR. The necessary values need to be provided when enabling the ruler. Document it in the thanos readme what and how to specify it, and then it needs to go in the plugindefinition.

viennaa (changes requested):
Thanks for pursuing the ruler component config 👍.

I think the cert handling is missing here, so we need to ensure to have config in place, to talk to an alertmanager.

Please also add `thanosruler.spec.routePrefix` to have it available at the `ruler` subpath after the query url.

thanos/charts/templates/ruler.yaml
trouaux: if configured, prometheus should already have the certs to communicate with alertmanager. ideally I would reuse those but in the mean time I can make it same setup as current prod
https://github.com/sapcc/helm-charts/blob/master/common/thanos/templates/ruler/alertmanager-config.yaml

viennaa: fair point. the ruler and query also could run in a cluster without prometheus, so then we would need this on top.

trouaux: i would have to look into this more, but if alertmanager is deployed we should be able to fetch TLS secrets and generate config for thanos. let's discuss this offline, i will add the configmap to templates for now which will require some user input when deploying the plugin

thanos/charts/templates/ruler/alertmanager-config.yaml
```diff
@@ -0,0 +1,11 @@
+{{ if .Values.thanos.ruler.enabled }}
```
viennaa, line 1: maybe we can make this config optional. Like `.Values.thanos.ruler.amconfig.enabled`

It defaults to false and expects the AM to be present in the cluster. Should have a not in the `values.yaml` to point out that this needs to be specified, if there is no AM around.

richardtief (approved):
Thanks! Let's ship it.

PR #542, opened 2024-12-10
[openTelemetry] Bumps version to v0.114.0

olandr:
- **chore(openTelemetry): bumps version of charts to v.114.0**
- **fix(openTelemetry): changes default tag for collectorImage to be 'main'**


timojohlo (approved):
Thanks for taking this!

PR #553, opened 2024-12-12
feat(kube-monitoring): Add Pod service discovery

### LINKED ISSUE #552: [kube-monitoring] Add service discovery for pods
## Context

The Greenhouse kube-monitoring plugin should provide an easy way to automatically detect scrape targets in the form of Kubernetes Pods if they have a certain label. This should listen to the port `metrics`.

## Tasks

- [x] Create a PodMonitor which implements the discovery via label
- [x] Test and verify
- [x] Documentation

richardtief:
Fixes #552


ibakshay (approved):
LGTM! 🚀

kube-monitoring/charts/values.yaml
```diff
@@ -197,6 +197,71 @@ kubeMonitoring:
...
+        any: true
+
+        ## Explicit list of namespace names to select
+        ##
+        # matchNames: []
+
+      ## Example which selects all Pods to be monitored
+      ## with label "monitoredby" with values any of "example-pod-1" or "example-pod-2"
+      selector:
+        matchExpressions: []
```
viennaa, line 254: The description is a copy paste error I assume.

Either we copy the [original](https://github.com/prometheus-community/helm-charts/blob/75bdfa0ed933adeacc452dabf3d40fbb01ba7cf1/charts/kube-prometheus-stack/values.yaml#L4426-L4434) or I would comment it out.

```suggestion
      ## Custom definition to pick up specific pods with an expression.
      # selector:
      ##  matchExpressions: []
```

kube-monitoring/charts/values.yaml
```diff
@@ -197,6 +197,71 @@ kubeMonitoring:
...
+
+        ## label selector for pods
+        ##
+        matchLabels:
+          greenhouse/scrape: "true"
```
viennaa, line 259: Do we really want to add this label everywhere needed or put it the other way around and add some `do-not-scrape` style?

viennaa (commented):
Thanks for resuming from the discussion. Let's discuss inline.

kube-monitoring/charts/values.yaml
richardtief: I would prefer an opt-in. Otherwise, we cannot ensure that multiple pods are scraped.

kube-monitoring/charts/values.yaml
richardtief: This leads to a failed linting.  What is the error here?

kube-monitoring/README.md
```diff
@@ -124,6 +124,10 @@ Greenhouse regularly performs integration tests that are bundled with **kube-mon
 | `alerts.alertmanager.tlsConfig.cert`          | TLS certificate for communication with Alertmanager                                                                 | `Secret`                 |
 | `alerts.alertmanager.tlsConfig.key`           | TLS key for communication with Alertmanager                                                                         | `Secret`                 |
 
+## Service Discovery
+
+The **kube-monitoring** Plugin uses a PodMonitor to automatically discover the Prometheus metrics of the Kubernetes Pods in any Namespace. The PodMonitor is configured to detect the metrics endpoint of the pods with the port name `metrics` **and** the label `greenhouse/scrape: “true”`.
```
viennaa, line 129:
```suggestion
The **kube-monitoring** Plugin uses a PodMonitor to automatically discover the Prometheus metrics of the Kubernetes Pods in any Namespace. The PodMonitor is configured to detect the metrics endpoint of the pods with the port name `metrics` **and** the label `greenhouse/scrape: “true”`.

*Important*: The label needs to be added manually to have the pod scraped and the port name needs to match.
```

kube-monitoring/charts/values.yaml
viennaa: Ok, let's see how it works out

kube-monitoring/charts/values.yaml
viennaa: Haven't run it, but then let's leave it in with a changed description.

```suggestion
      ## Custom definition to pick up specific pods with an expression.
      selector:
        matchExpressions: []
```

viennaa (approved):
Thanks for tackling it :)

Issue #515, opened 2024-11-19
[opentelemetry] Templating Log-Parsing


timojohlo:
## Context 
We want to offer log-parsing packages for common scenarios such as KVM or Ceph. For these scenarios users can decided what parsing they need and tick the corresponding boxes in the PluginDefinition to enable them.

For this to work we will template different logging scenarios, build if/else templating for each scenario and outsource each scenario to their own file, which should result in a better-structured config file for the logs-collector.

## Acceptance criteria
- [ ] Template KVM logging
- [ ] Template Ceph logging

PR #557, opened 2024-12-17
[opentelemetry] add templating for ceph and kvm

timojohlo:
## Pull Request Details

This PR adds a functionality to enable/disable common log configuration for logs for Ceph and KVM.
Both logtypes can be disabled/enabled in the PluginDefinition. In case of enabling, the logs-collector.yaml will be enriched by KVM/Ceph specific labels, parsing and defined pipelines.

The goal is to reduces the complexity of the logs-collector configuration and provide standard log configuration for common scenarios.

Corresponding Roadmap issue: #515

### DIFF
--- a/opentelemetry/chart/templates/_kvm-config.tpl
+++ b/opentelemetry/chart/templates/_kvm-config.tpl
@@ -1,0 +1,53 @@
+{{- define "kvm.labels" }}
+
+{{- end }}
+{{- define "kvm.receiver" }}
+filelog/kvm_logs:
+  include: [ /var/log/libvirt/qemu/*.log, /var/log/openvswitch/*.log ]
+  include_file_path: true
+  start_at: beginning
+  multiline:
+    line_start_pattern: ^\d{4}-\d{2}-\d{2}
+  operators:
+    - type: regex_parser
+      regex: (?P<logtime>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3})
+    - type: time_parser
+      parse_from: attributes.logtime
+      layout: '%Y-%m-%dT%H:%M:%S.%L'
+      layout_type: strptime
+    - id: file-label
+      type: add
+      field: attributes["log.type"]
+      value: "files"
+{{- end }}
+{{- define "kvm.transform" }}
+transform/kvm_openvswitch:
+  error_mode: ignore
+  log_statements:
+    - context: log
+      conditions:
+        - resource.attributes["k8s.daemonset.name"] == "neutron-openvswitch-agent"
+      statements:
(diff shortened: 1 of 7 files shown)


opentelemetry/chart/templates/_kvm-config.tpl
```diff
@@ -0,0 +1,64 @@
+{{- define "kvm.labels" }}
+
+{{- end }}
```
viennaa, line 3:
```suggestion
```

Is this needed?

opentelemetry/chart/templates/logs-collector.yaml
```diff
@@ -208,33 +194,11 @@ spec:
...
-              - merge_maps(attributes, ExtractGrokPatterns(body, "%{GREEDYDATA:osd.wall.type}\\:%{SPACE}%{NUMBER:osd.wall.writes:float}(K|M|G)?%{SPACE}writes,%{SPACE}%{NUMBER:osd.wall.syncs:float}(K|M|G)?%{SPACE}syncs,%{SPACE}%{NUMBER:osd.wall.writes_per_sync:float}%{SPACE}writes per sync, written\\:%{SPACE}%{NUMBER:osd.wall.written_gb:float}%{SPACE}(GB|MB|TB)?,%{SPACE}%{NUMBER:osd.written_mb_sec:float}", true),"upsert")
-              - merge_maps(attributes, ExtractGrokPatterns(body, "%{WORD:log_level}%{SPACE}%{NOTSPACE}%{SPACE}%{SPACE}%{NOTSPACE:process.name}", true),"upsert")
-              - set(attributes["config.parsed"], "ceph_osd") where attributes["osd.stats.level"] != nil
-              - set(attributes["config.parsed"], "ceph_osd") where attributes["log_level"] != nil
+              - set(attributes["network.protocol.name"], ConvertCase(attributes["network.protocol.name"], "lower")) where attributes["network.protocol.name"] != nil
+              - set(attributes["config.parsed"], "ingress-nginx") where attributes["client.address"] != nil
```
viennaa, line 198: What is this needed for?

viennaa (commented):
I like the splitting of the main file to make it more clean and easier to use.

Two minor things inside.

opentelemetry/chart/templates/_kvm-config.tpl
timojohlo: It is not needed at the moment. I left it there as we will add KVM labels in the future, but I agree that we can remove it for now.

viennaa: Just add it when the feature is popping up :)

opentelemetry/chart/templates/logs-collector.yaml
timojohlo: The value for network.protcol.name is (if filled) HTTP or http, so we want to be consistent regarding upper/lower-case.
The value for client.address (if filled) is usually an IP address.  Here we assign that it was parsed with the transform/ingress pipeline if the IP address has been set. 
That said, both lines are subject to change but currently allow us to trace different pipelines based on a defined condition.

opentelemetry/chart/templates/_kvm-config.tpl
timojohlo: Agree, thank you!

viennaa (approved):

Issue #546, opened 2024-12-10
🐛 [BUG] - Alerts Plugin regenerates CA on every diff


IvoGoman:
### Priority

(Medium) I'm annoyed but I'll live

### Description

The diff for the alerts plugin shows a diff for the `ccloud-monitoring-ca` and `ccloud-ca-bundle` on every reconciliation. 

```text
"Secret/ccloud-ca-bundle: [\n {\n \"value\": \"***** - after\",\n \"op\": \"replace\",\n \"path\": \"/data/ca.crt\"\n }\n],Secret/ccloud-monitoring-ca: [\n {\n \"value\": {\n \"apiVersion\": \"v1\",\n \"data\": {\n \"tls.crt\": \"*****\",\n \"tls.key\": \"*****\"\n },\n \"kind\": \"Secret\",\n \"metadata\": {\n \"annotations\": {\n \"helm.sh/hook\": \"pre-install, pre-upgrade\"\n },\n \"labels\": {\n \"app.kubernetes.io/managed-by\": \"Helm\",\n \"plugin\": \"alerts-ccloud\",\n \"plugindefinition\": \"alerts\",\n \"release\": \"alerts-ccloud\"\n },\n \"name\": \"ccloud-monitoring-ca\",\n \"namespace\": \"ccloud\"\n },\n \"type\": \"Opaque\"\n },\n \"op\": \"add\",\n \"path\": \"\"\n }\n]"}
```

The `ccloud-monitoring-ca` should not be regenerated on every deploy, but the condition does not work:
https://github.com/cloudoperators/greenhouse-extensions/blob/main/alerts/charts/templates/ca-secret-issuer-cert.yaml#L2

This causes reoccurring `helm upgrades` which eventually lead to the release stuck in pending. 

### Reproduction steps

```bash
1. Go to the organisation namespace 
2. `helm history alerts`
3. see frequent upgrades of the plugin

5239    	Tue Dec 10 17:43:41 2024	superseded	alerts-0.15.3	           	2.3.3
5240    	Tue Dec 10 17:44:49 2024	superseded	alerts-0.15.3	           	2.3.3
5241    	Tue Dec 10 17:45:55 2024	superseded	alerts-0.15.3	           	2.3.3
5242    	Tue Dec 10 17:47:02 2024	superseded	alerts-0.15.3	           	2.3.3
5243    	Tue Dec 10 17:48:08 2024	deployed  	alerts-0.15.3	           	2.3.3
```
```


[...]

Issue #550, opened 2024-12-11
[alerts] Improve automatic CA generation and validity


richardtief:
## Context

The Alerts Plugin creates a CA certificate to authenticate subsequent Prometheus installations that send alerts to the Alertmanager. This process needs to be redesigned.

## Tasks

- [x] improve the CA creation process to enable multiple options
- [x] test and verify
- [x] documentation

PR #559, opened 2024-12-18
feat(alerts): improved ca cert generation

richardtief:
fixes: #546, #550


alerts/README.md
```diff
@@ -205,6 +214,14 @@ alertmanagerConfiguration:
   name: global-alertmanager-configuration
 ```
 
+## TLS Certificate Requirement
+
+In order for Prometheus installations in onboarded Greenhouse clusters to communicate with the Alertmanager component, the Ingress requires a TLS certificate for which Prometheus is configured as trusted. There are various ways in which you can generate/configure the required TLS certificate.
```
viennaa, line 219: Hard to wrap it in words. Tried to make it a little bit more explanatory, especially the first sentence. 

```suggestion
Greenhouse onboarded Prometheus installations need to communicate with the Alertmanager component to enable advanced processing of alerts. The Alertmanager Ingress requires a TLS certificate to be configured and trusted by Prometheus to ensure the communication. There are various ways in which you can generate/configure the required TLS certificate.
```

alerts/README.md
```diff
@@ -205,6 +214,14 @@ alertmanagerConfiguration:
...
+
+  - You can use an automatically generated self-signed certificate by setting `alerts.auth.autoGenerateCert.enabled` to `true`. Helm will create a self-signed cert and a secret for you.
+  - You can use your own generated self-signed certificate by setting both `alerts.auth.autoGenerateCert.enabled` to `false`. You should provide the necessary values to `alerts.auth.certFile`, `alerts.auth.keyFile`, and `alerts.auth.caFile`.
```
viennaa, line 222: what do you mean with `both` here? You are just referring to `alerts.auth.autoGenerateCert.enabled` or am I missing something?

viennaa (commented):
Cool, like it. Little things inline.

Haven't tried cert generation but looks sane to me and I am assuming, you have templated this before.

viennaa:
It just crossed my mind. Could we actually test this? Thinking of having the certs created and then we check, if they are really valid?

alerts/README.md
richardtief: There is no second option. I'm going to remove the word `both`
```suggestion
  - You can use your own generated self-signed certificate by setting `alerts.auth.autoGenerateCert.enabled` to `false`. You should provide the necessary values to `alerts.auth.certFile`, `alerts.auth.keyFile`, and `alerts.auth.caFile`.
```

alerts/charts/templates/tests/test-alerts-config.yaml
```diff
@@ -61,4 +48,18 @@ data:
...
+    {{- if and .Values.alerts.auth.autoGenerateCert.enabled .Values.alerts.alertmanager.ingress.enabled .Values.alerts.alertmanager.ingress.hosts (.Capabilities.APIVersions.Has "cert-manager.io/v1") }}
+    @test "Verify certificate creation and validation against the API endpoint of the alert manager" {
+        verify "there is 1 issuer named '{{ .Release.Namespace }}-monitoring-issuer'"
+        verify "there is 1 certificate named '{{ .Release.Namespace }}-prometheus-auth'"
+        verify "there is 1 secret named 'tls-{{ .Release.Namespace }}-prometheus-auth'"
+
+        url="https://{{ first .Values.alerts.alertmanager.ingress.hosts }}/-/healthy"
+
+        response=$(curl --cert /tls-assets/tls.crt --key /tls-assets/tls.key ${url})
+        [ "$response" == "OK" ]
```
viennaa, line 61: Suggesting to remove not needed var which looks more native to me according to various examples in the documentation. At least if we're not checking for a particular return string. OK should yield return code `0`.

```suggestion
        run curl --cert /tls-assets/tls.crt --key /tls-assets/tls.key ${url}
        [ "$status" -eq "0" ]
```

viennaa (commented):
little suggestion to the test syntax.

viennaa (approved):

PR #581, opened 2025-01-19
fix(plutono): update service URL format

ibakshay:
Adjust the Service URL format based on this PR https://github.com/cloudoperators/greenhouse/pull/840


viennaa (approved):
Thanks!

PR #585, opened 2025-01-22
chore(perses): bump version to 0.1.9 and update Perses chart and plugin versions

ibakshay:
Increment the versions of the Perses chart and plugin to 0.9.0.


timojohlo (approved):
lgtm

PR #593, opened 2025-01-28
chore(plutono) update plugin test image

ibakshay:
Change the test framework image configuration to use our maintained image.


richardtief (changes requested):
Thanks. Please increase the versions in `Chart.yaml `and `plugindefinition.yaml`.

ibakshay:
> Thanks. Please increase the versions in `Chart.yaml `and `plugindefinition.yaml`.

thank you. fixed in dfe58485a72f71b981796a306cc062f721c06839

richardtief (approved):
Thanks. Good catch

PR #616, opened 2025-02-03
feat(kube-monitoring): refine pod service discovery

richardtief:
Switch from label selector to annotation.

### DIFF
--- a/kube-monitoring/charts/templates/_helper.tpl
+++ b/kube-monitoring/charts/templates/_helper.tpl
@@ -85,7 +85,7 @@ release: {{ .Release.Name }}
 {{- end }}
 
 {{/* Generate basic labels */}}
-{{ define "kube-monitoring.labels" }}
+{{ define "kubeMonitoring.labels" }}
 {{- $path := index . 0 -}}
 {{- $root := index . 1 -}}
 plugindefinition: kube-monitoring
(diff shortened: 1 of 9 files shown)


trouaux (approved):

kube-monitoring/charts/templates/_helper.tpl
```diff
@@ -85,7 +85,7 @@ release: {{ .Release.Name }}
 {{- end }}
 
 {{/* Generate basic labels */}}
-{{ define "kube-monitoring.labels" }}
+{{ define "kubeMonitoring.labels" }}
```
ibakshay, line 88: Shall we name it as annotations? WDYT? 
```suggestion
{{ define "kubeMonitoring.annotations" }}
```

kube-monitoring/charts/templates/_helper.tpl
```diff
@@ -97,7 +97,7 @@ app.kubernetes.io/managed-by: {{ $root.Release.Service }}
 release: {{ $root.Release.Name | quote }}
 {{- end }}
 
-{{- define "kube-monitoring.dashboardSelectorLabels" }}
+{{- define "kubeMonitoring.dashboardSelectorLabels" }}
```
ibakshay, line 100:
```suggestion
{{- define "kubeMonitoring.dashboardSelectorAnnotations" }}
```

kube-monitoring/charts/templates/_helper.tpl
richardtief: No, those are still labels. 🙂

kube-monitoring/charts/templates/_helper.tpl
richardtief: Again, still labels. I just renamed the helper template.

kube-monitoring/charts/values.yaml
```diff
@@ -206,31 +206,27 @@ kubeMonitoring:
...
+      #   targetLabel: nodename
+      #   replacement: $1
+      #   action: replace
+
+      ## MetricRelabelConfigs to apply to samples after scraping, but before ingestion.
+      ## ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#relabelconfig
       ##
-      podMetricsEndpoints:
-          ## Monitor Pods with the following port name
-        - port: metrics
```
viennaa, line 214: Why did you want to move this?

viennaa (commented):
Thanks for pushing this.

Could you elaborate a little more why this is needed? I can spot the relabelings you've added to `podMeticsEndpoints` but apart from that I can't see the mentioned annotations change.

kube-monitoring/charts/values.yaml
richardtief: The annotations selection is a bit hidden. https://github.com/cloudoperators/greenhouse-extensions/blob/2b01d25353046ddd90849b036a8f433297fdbd76/kube-monitoring/charts/templates/podmonitor-sd.yaml#L15

viennaa: ok then I've parsed it correctly 👍

PR #625, opened 2025-02-05
fix(thanos) fix svc label selector

trouaux:


thanos/charts/templates/ruler/ruler-svc.yaml
```diff
@@ -26,5 +26,5 @@ spec:
     protocol: TCP
     targetPort: http
   selector:
-    app.kubernetes.io/managed-by: {{ include "release.name" . }}
-    app.kubernetes.io/name: ruler
+    app.kubernetes.io/managed-by: prometheus-operator
```
viennaa, line 29: Do we actually need to select this? The bottom line should be sufficient, no?

trouaux: won't that also select unwanted pods managed by prometheus-operator ?

viennaa (approved):
Thanks!

Issue #462, opened 2024-10-24
[kube-monitoring] Show alerts about Prometheus/Thanos issues


artherd42:
### Priority

(Medium) I'm annoyed but I'll live

### Is your feature request related to a problem?

Operation teams should "see" when a Prometheus/Thanos is not healthy anymore.

### Acceptance Criteria

- [x] The Alert Dashboard shows the alert(s).

richardtief:
We're going to leverage the blackbox-exporter in the greenhouse namespace to probe ingresses and static targets. The alerts can then be send to Alertmanagers in the respective org namespace.

trouaux:
https://github.com/cloudoperators/greenhouse-extensions/pull/937
https://github.com/cloudoperators/greenhouse-extensions/pull/938
https://github.com/sapcc/helm-charts/pull/8901

trouaux:
grpc_prober has been added to blackbox config in kube-monitoring plugin
thanos plugin can create ServiceMonitor to probe it's store list

we are discussing where thanos global (central cluster) will be deployed

trouaux:
allow central cluster thanos/kube-monitoring plugin install https://github.com/cloudoperators/greenhouse/pull/1238

PR #630, opened 2025-02-06
feat(kube-monitoring): adds prometheus blackbox-exporter

richardtief:
The blackbox exporter can be utilised to read Probe CRs. A Probe defines how groups of ingresses and static targets should be monitored.

foundation to: #462


ibakshay (approved):
LGTM! 🚀
I also learned about blackbox_exporter now :)

viennaa (commented):
I like the idea. Do you want to ship probes with the respective chart and check for the capability of the CR?

richardtief:
The first use case is within the Greenhouse central cluster probing the Prometheus and Thanos urls if they are up and running.

viennaa (approved):

trouaux (approved):

PR #633, opened 2025-02-07
feat(thanos) ServiceMonitor

trouaux:
`  serviceMonitorSelector:
    matchLabels:
      plugin: kube-monitoring-obs-eu-de-1`

@richardtief prometheus servicemonitor selector contains release name, is there a way to pass it to other plugins before creation ?

needed for servicemonitor labeling


richardtief:
There are two ways to do this: 

1. you leave the configuration open, as opentelemetry does:
   - https://github.com/cloudoperators/greenhouse-extensions/blob/099db492e50d246c37714489750e46e390ca7064/opentelemetry/chart/values.yaml#L65
   - https://github.com/sapcc/helm-charts/blob/a15ec624ab5a38ffa933323cb2d8d7fabc6f44fd/system/greenhouse-ccloud/templates/opentelemetry-pluginpreset.yaml#L41:L43
2. use the pod service discovery implemented in [kube-monitoring] (https://github.com/cloudoperators/greenhouse-extensions/blob/main/kube-monitoring/README.md#service-discovery)

Both solutions require the release name of the Prometheus that should pick up the ServiceMonitor. The requirement is that the Thanos components belonging to the Prometheus setup are collected by the same Prometheus.

trouaux:
> 2. use the pod service discovery implemented in [kube-monitoring] (https://github.com/cloudoperators/greenhouse-extensions/blob/main/kube-monitoring/README.md#service-discovery)
good to know about that one, in this case since there is multiple pods feels easier to create a ServiceMonitor

if user inputs prometheus release.name during plugin creation we can do 1

thanos/charts/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and Greenhouse contributors
+# SPDX-License-Identifier: Apache-2.0
```
richardtief, line 2: Please remove those headers. They will prevent Helm from templating if the file is empty.

thanos/charts/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+{{ if .Values.thanos.serviceMonitorLabels }}
```
richardtief, line 4: Cool. But I would suggest 
```suggestion
{{ if .Values.thanos.selfMonitor.enabled }}
```

thanos/charts/values.yaml
```diff
@@ -19,6 +19,8 @@ thanos:
   # used across the stack
   httpAddress:
 
+  serviceMonitorLabels:
```
richardtief, line 22:
```suggestion
  # Deploys a ServiceMonitor to self-monitor the Thanos components
  selfMonitor:
    enabled: true
```

thanos/plugindefinition.yaml
```diff
@@ -42,6 +42,10 @@ spec:
           name: thanos.query.autoDownsampling
           required: false
           type: bool
+        - description: Needed to create a ServiceMonitor for Thanos to expose metrics to Prometheus
+          name: thanos.serviceMonitorLabels
```
richardtief, line 46:
```suggestion
          name: thanos.selfMonitor.enabled
```

thanos/charts/templates/tests/test-thanos-config.yaml
```diff
@@ -47,10 +47,16 @@ data:
...
-        try "at most 3 times every 5s to get pods named 'thanos-ruler-{{ .Release.Name }}' and verify that '.status.phase' is 'running'"
-        try "at most 3 times every 5s to get configmaps named '.*(ruler)+.*(rulefiles)+.*' and verify that '.data' matches '.*yaml.*'"
+      verify "there is 1 statefulset named 'thanos-ruler-{{ .Release.Name }}'"
+      verify "there is 1 service named 'thanos-ruler-operated'"
+      try "at most 3 times every 5s to get pods named 'thanos-ruler-{{ .Release.Name }}' and verify that '.status.phase' is 'running'"
+      try "at most 3 times every 5s to get configmaps named '.*(ruler)+.*(rulefiles)+.*' and verify that '.data' matches '.*yaml.*'"
+    }
+    {{ end }}
+
+    {{ if .Values.thanos.serviceMonitorLabels }}
```
richardtief, line 57:
```suggestion
    {{ if .Values.thanos.selfMonitor.enabled }}
```

richardtief (changes requested):
Thanks. Just one issue which will occur and a suggestion.

thanos/charts/templates/servicemonitor.yaml
trouaux: I see your point, I would go for how kube-prometheus-stack does it then
  
serviceMonitor:
    ## If true, create a serviceMonitor for prometheus operator
    ##
    selfMonitor: true

    ## Labels for ServiceMonitor
    additionalLabels: {}

richardtief (approved):
Thanks. Let's ship it.

PR #639, opened 2025-02-13
feat(thanos) automate Thanos version update with renovate

trouaux:
removed appVersion from **Chart.yaml**

and using **values.yaml** `thanos.image.tag` as default thanos version

this will enable renovate to automatically update thanos version when released


.github/renovate.json
```diff
@@ -43,6 +43,19 @@
...
       "extractVersionTemplate": "^(?<version>.*)$"
+    },
+    {
+      "customType": "regex",
+      "description": "Bump thanos version in the plugindefinition",
+      "fileMatch": [
+        "^thanos.charts.Chart.yaml$"
+      ],
+      "matchStrings": [
+        "appVersion:\\s*(?<currentValue>v?(?:\\d+\\.){0,2}\\d+)"
```
ibakshay, line 54: What are you trying to achieve? Do you want to bump the appVersion of Thanos when there is a new release in [thanos-io/thanos](https://github.com/thanos-io/thanos/releases)?

kengou (changes requested):
This is not necessary, renovate is picking it up automatically. The custom manager is only needed for plugindefinitions to be automatically updated.

trouaux:
> This is not necessary, renovate is picking it up automatically. The custom manager is only needed for plugindefinitions to be automatically updated.

everything under dependencies in Chart.yaml seems to be picked up but not the `appVersion` field, not sure how it could fetch the version since we use it as a template variable

kengou:
> @kengou
> 
> > This is not necessary, renovate is picking it up automatically. The custom manager is only needed for plugindefinitions to be automatically updated.
> 
> everything under dependencies in Chart.yaml seems to be picked up but not the `appVersion` field, not sure how it could fetch the version since we use it as a template variable

It should be picked up like here: https://github.com/cloudoperators/greenhouse-extensions/pull/649/files
Maybe you need to add [thanos-io/thanos](https://github.com/thanos-io/thanos/releases) as a dependency

richardtief:
It's better to move the appVersion field to the values file. If it's maintained as a ``, it works. e.g. https://github.com/cloudoperators/greenhouse-extensions/pull/608/files

trouaux:
> It's better to move the `appVersion` field to `values.yaml`. If it's maintained as a `tag`, it works. e.g. https://github.com/cloudoperators/greenhouse-extensions/pull/608/files

we remove it from templates too ? `{{ .Values.thanos.image.repository }}:{{ .Values.thanos.image.tag | default .Chart.AppVersion }}`

richardtief:
If you wish ‘renovate’ to take over, yes. 
> we remove it from templates too ? `{{ .Values.thanos.image.repository }}:{{ .Values.thanos.image.tag | default .Chart.AppVersion }}`

kengou (commented):
ok from my side

kengou (approved):

richardtief (approved):

Issue #589, opened 2025-01-23
[plugin-chart-test] Improve Helm Test Cleanup and Logging


ibakshay:
### Priority

None

### Is your feature request related to a problem?

Currently, when a Helm chart test fails:
1. Resources created by the test are not immediately cleaned up, leaving debris in the cluster
2. Logs from the test pod are not added to the plugin resource status field. This forces users to manually run `helm test` in remote clusters to debug test failures, leading to delays in troubleshooting.

## **Acceptance Criteria**

- [ ] **Test Pod Logs in Plugin CR Status Field**:
   - Logs from the Helm test pod are captured upon test completion and appended to the `status` field of the plugin custom resource (CR).
   - Logs are concise, well-formatted, and user-friendly for quick debugging.
   - Users can view the test logs directly in the plugin CR status without manually interacting with the cluster.

- [x]  **Resource Cleanup**:
   - [x] All resources created during a Helm test are cleaned up after the plugin is removed. 
   - [x] Cleanup logic addresses ConfigMaps, Pods, RBAC permissions, and any other resources created during tests.

- [x] **Update Documentation** 
   - [ ] documentation is updated accordingly

### Reference Issues

Depends on  https://github.com/cloudoperators/greenhouse/issues/585

ibakshay:
This issue can be tackled soon after this issue is closed - https://github.com/cloudoperators/greenhouse/issues/585

ibakshay:
@richardtief and myself aligned on introducing a separate status field in the plugin for the helm chart test results. 
the status field will look like below.
```yaml
 status:
  helmChartTestSucceeded:
    lastTransitionTime: "2025-02-13T21:28:47Z"
    status: "False"
    conditions: 
     - message: Verify that there is a service named perses
       succeeded: true
     - message: Verify that there is a configmap named perses
       succeeded: true
     - message: Verify failure deployment and running status of the perses pod
       succeeded: false
````
 @IvoGoman Do you have any comments on this structure? 
This is an enhancement to this issue https://github.com/cloudoperators/greenhouse/issues/106

IvoGoman:
[...]

PR #656, opened 2025-02-18
chore(plugins): remove test annotations

ibakshay:
Previously, test annotations were applied to multiple resources, but these resources were neither cleaned up when a test failed nor when the plugin was uninstalled. This led to orphaned resources persisting beyond their intended lifecycle.

With this change, these resources will be part of the plugin install/uninstall lifecycle. 

relates to https://github.com/cloudoperators/greenhouse-extensions/issues/589


richardtief (approved):
As discussed, this looks good. Thanks!

PR #661, opened 2025-02-21
chore(deps): upgrade chart kube-prometheus-stack to v69.3.3

richardtief:
To use the latest feature:
https://github.com/prometheus-community/helm-charts/pull/5339


ibakshay (approved):
Awesome work on contributing to upstream!! 🚀🔥
Non-blocking suggestion: Does it make sense to add any test for this use case? Is it possible

richardtief:
> Awesome work on contributing to upstream!! 🚀🔥 Non-blocking suggestion: Does it make sense to add any test for this use case? Is it possible

Sure, we will use this feature and then test it properly.

Issue #142, opened 2024-06-14
[kube-monitoring] Only Prometheus deployment function


richardtief:
## Context 

kube-monitoring plugin deploys a collection of Kubernetes manifests by default. To scale horizontally with multiple Prometheus in a cluster, the a new `PluginDefinition` should be created to be able to deploy only Prometheus. The chart preferably is the same as kube-monitoring. 

## Acceptance criteria

1. Ensure prometheus-operator + CRDs can be installed independently - or make it a prerequisite
2. Test 
3. Documentation

richardtief:
We are going to create a own prometheus plugin based on the kube-monitoring helm chart.

PR #662, opened 2025-02-21
feat(kube-monitoring): ensure unique secret names based on the release

richardtief:
To enable multiple deployments in the same namespace. Foundation to: #142

### DIFF
--- a/kube-monitoring/charts/templates/_alertmanager-config.yaml.tpl
+++ b/kube-monitoring/charts/templates/_alertmanager-config.yaml.tpl
@@ -2,8 +2,13 @@
   scheme: https
 {{- if and .Values.alerts.enabled .Values.alerts.alertmanagers.hosts }}
   tls_config:
-    cert_file: /etc/prometheus/secrets/tls-prometheus-alertmanager-auth/tls.crt
-    key_file: /etc/prometheus/secrets/tls-prometheus-alertmanager-auth/tls.key
+{{- if and .Values.alerts.alertmanagers.tlsConfig.cert .Values.alerts.alertmanagers.tlsConfig.key }} 
+    cert_file: /etc/prometheus/secrets/tls-prometheus-{{ .Release.Name }}/tls.crt
+    key_file: /etc/prometheus/secrets/tls-prometheus-{{ .Release.Name }}/tls.key
+{{- else }}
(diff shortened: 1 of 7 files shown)


kube-monitoring/charts/templates/_alertmanager-config.yaml.tpl
```diff
@@ -2,8 +2,13 @@
...
   tls_config:
-    cert_file: /etc/prometheus/secrets/tls-prometheus-alertmanager-auth/tls.crt
-    key_file: /etc/prometheus/secrets/tls-prometheus-alertmanager-auth/tls.key
+{{- if and .Values.alerts.alertmanagers.tlsConfig.cert .Values.alerts.alertmanagers.tlsConfig.key }} 
+    cert_file: /etc/prometheus/secrets/tls-prometheus-{{ .Release.Name }}/tls.crt
+    key_file: /etc/prometheus/secrets/tls-prometheus-{{ .Release.Name }}/tls.key
+{{- else }}
+    cert_file: /etc/prometheus/secrets/tls-prometheus-{{ .Release.Namespace }}/tls.crt
+    key_file: /etc/prometheus/secrets/tls-prometheus-{{ .Release.Namespace }}/tls.key
+{{- end }}
```
viennaa, line 11: This would always point to the dummy-secret, wouldn't it?

In this case nothing would work, but the at least the operator wouldn't complain. This captures the case of a prometheus without an alertmanager then, is this something we need?

kube-monitoring/charts/values.yaml
```diff
@@ -353,11 +354,13 @@ kubernetes-operations:
       - name: plugin
         value: "{{ $.Release.Name }}"
 
-
+## Configures Prometheus Alertmanager
 alerts:
   enabled: false
+  dummySecret: false
```
viennaa, line 360: Add a comment indicating its use please.

viennaa (commented):
Thanks for preparing this more generic approach.

I need some `dummySecret` explanation though.

kube-monitoring/charts/templates/_alertmanager-config.yaml.tpl
richardtief: The `dummySecret` is just for the github-action as the alerts plugin is a dependency.

Issue #592, opened 2025-01-27
[opentelemetry] add enable/disable feature for failover connector


timojohlo:
## Context
The [failover connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/failoverconnector) will be implemented with #591. In the future users should be able to enable/disable the failover feature. In case the opt out only one secret for username and password is necessary.

PR #685, opened 2025-02-28
feat(opentelemetry) add enable/disable for failover

timojohlo:
## Pull Request Details

Implements #592: allows users to enable/disable the failover-mechanism in the Plugin Definition

### DIFF
--- a/opentelemetry/README.md
+++ b/opentelemetry/README.md
@@ -82,7 +82,7 @@ The **OpenTelemetry** Plugin comes with a [Failover Connector](https://github.co
 `openTelemetry.logsCollector.enabled`    | Activates the standard configuration for logs | bool | `false` |
 `openTelemetry.logsCollector.kvmConfig.enabled`    | Activates the configuration for KVM logs (requires logsCollector to be enabled) | bool | `false` |
 `openTelemetry.logsCollector.cephConfig.enabled`    | Activates the configuration for Ceph logs (requires logsCollector to be enabled) | bool | `false` |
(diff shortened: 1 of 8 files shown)


opentelemetry/README.md
```diff
@@ -82,11 +82,12 @@ The **OpenTelemetry** Plugin comes with a [Failover Connector](https://github.co
 `openTelemetry.logsCollector.enabled`    | Activates the standard configuration for logs | bool | `false` |
 `openTelemetry.logsCollector.kvmConfig.enabled`    | Activates the configuration for KVM logs (requires logsCollector to be enabled) | bool | `false` |
 `openTelemetry.logsCollector.cephConfig.enabled`    | Activates the configuration for Ceph logs (requires logsCollector to be enabled) | bool | `false` |
+`openTelemetry.logsCollector.failover.enabled`      | Activates the failover mechanism  for shipping logs via a fallback user if the first users' credentials have expired | bool | `false` |
```
olandr, line 85: "First user" maybe you could highlight that it is intended to be `failover_a`? From an outsider's perspective it might be unclear which is the "first" and which is the "secondary".

olandr (changes requested):
Looks good, just a minor thing with how we document this feature. See my comment. In general, the idea that `failover.enabled` means that `failover_b` is active is a bit confusing considering `failover_a` is always active and the primary user.


So maybe we need to be careful in saying that `failover_a` is the primary one and is always present.

olandr (approved):

Issue #690, opened 2025-03-04
Automate `helm-docs` Validation on `values.yaml` Changes


ibakshay:
To ensure that README.md file for the particular plugin is properly updated, we want to set up a GitHub Action that checks whether `make generate-readme PLUGIN=<PLUGIN-NAME>` has been run after modifying `values.yaml`. This action should **not** automatically commit changes but should fail the workflow with a PR comment if the user has not updated the documentation.

### Plugins to Track

- [ ] alerts
- [ ] kube-monitoring
- [ ] opentelemetry
- [ ] perses
- [ ] plutono
- [ ] prometheus
- [ ] thanos

### Expected Behavior

1. A user modifies `values.yaml` in any of the above plugins.
2. The GitHub Action runs and verifies if `make generate-readme PLUGIN=<PLUGIN-NAME>` has been executed.
3. If the documentation is outdated, the workflow fails and informs the user by posting a PR review comment  to run the command manually.
4. If the documentation is up to date, the workflow passes.

### Implementation Plan

- Set up a GitHub Action that:
  - Triggers on changes to `values.yaml` in any plugin directory.
  - Runs `make generate-readme PLUGIN=<PLUGIN-NAME>`.
  - Checks for uncommitted changes in the repository (`git diff --exit-code`).
  - Fails if changes are detected, prompting the user to update the documentation.
---
**Related Resources:**
- [helm-docs GitHub Repository](https://github.com/norwoodj/helm-docs)

PR #692, opened 2025-03-04
feat(repo): automate helm-docs Validation on values.yaml changes

ibakshay:
This PR is the foundation for resolving https://github.com/cloudoperators/greenhouse-extensions/issues/690


.github/workflows/helm-docs.yaml
```diff
@@ -0,0 +1,164 @@
...
+      - "**/*/*/values.yaml"
+
+jobs:
+  helm-docs-check:
+    runs-on: [default]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
kengou, line 16: Do you really need all history for all tags and branches ?

.github/workflows/helm-docs.yaml
```diff
@@ -0,0 +1,164 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up go environment
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22.0
```
kengou, line 22: I would use the latest stable go version and not an old one 
```suggestion
        with:
          go-version: 'stable'
```

.github/workflows/helm-docs.yaml
```diff
@@ -0,0 +1,164 @@
+name: "Helm Docs Check"
+
+on:
+  pull_request:
+    paths:
+      - "**/*/values.yaml"
+      - "**/*/*/values.yaml"
```
kengou, line 7: If you want the action be triggered once the PR is updated you need to add the `types: [opened, synchronize, reopened]` as a trigger. Otherwise it will run only once when the PR is opened
```suggestion
  pull_request:
    types: [opened, synchronize, reopened]
    paths:
      - "**/*/values.yaml"
      - "**/*/*/values.yaml"
```

.github/workflows/helm-docs.yaml
ibakshay: Yes, I require as I am doing `git diff` [here](https://github.com/cloudoperators/greenhouse-extensions/blob/ibakshay/validate-helm-docs-gha/.github/workflows/helm-docs.yaml#L29)

.github/workflows/helm-docs.yaml
ibakshay: Thank you. Fixed in e149e6eb5b38a27ff1246bd0ba852e48cc7ea754

.github/workflows/helm-docs.yaml
ibakshay: thank you for the pointer. Fixed in e149e6eb5b38a27ff1246bd0ba852e48cc7ea754

PR #695, opened 2025-03-05
feat(thanos) add ingress for query service

trouaux:
service-proxy doesn't support grpc endpoints and currently it is not possible to expose query frontend as a custom domain

to allow our central/global query to pull data from our remote clusters and to allow to expose central query using the global domain we need to create ingresses


thanos/charts/templates/query/ingress-grpc.yaml
```diff
@@ -0,0 +1,40 @@
...
+    {{- range .Values.thanos.query.ingress.grpc.hosts }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+          {{- range .paths }}
+          - path: {{ .path }}
+            pathType: {{ .pathType }}
+            backend:
+              service:
+                name: {{ include "release.name" . }}-query
```
richardtief, line 25: This will not work. In the scope of the range you can work with `$` to include templates.
```suggestion
                name: {{ include "release.name" $ }}-query
```

thanos/charts/templates/query/ingress.yaml
```diff
@@ -0,0 +1,40 @@
...
+    {{- range .Values.thanos.query.ingress.hosts }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+          {{- range .paths }}
+          - path: {{ .path }}
+            pathType: {{ .pathType }}
+            backend:
+              service:
+                name: {{ include "release.name" . }}-query
```
richardtief, line 25: Same here.
```suggestion
                name: {{ include "release.name" $ }}-query
```

thanos/charts/values.yaml
```diff
@@ -72,6 +72,62 @@ thanos:
...
+        # -- Additional annotations for the Ingress resource.(GRPC) To enable certificate autogeneration, place here your cert-manager annotations.
+        # For a full list of possible ingress annotations, please see
+        # ref: https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/annotations.md
+        annotations: {}
+
+        # -- Ingress TLS configuration. (GRPC)
+        tls: []
+        #    - secretName: ingress-cert
+        #      hosts: []
+
```
richardtief, line 130: The indentation is slightly off, right?. It's `thanos.query.ingress.grpc` instead of `thanos.query.grpc`

richardtief (changes requested):
Thank you! I really like the quality of the manifests. It's easy to read. The same applies for the helm docs rendering. 👍🏻

thanos/charts/values.yaml
trouaux: still brainstorming what would be the best here, current setup is `ingress` contains `grpc` for an extra ingress resource focusing on grpc. so this is intended `thanos.query.ingress.grpc `instead of `thanos.query.grpc`

but i'm considering only 1 ingress resource that we could configure for both grpc and http

PR #721, opened 2025-03-19
feat(kube-monitoring) use serviceName option

trouaux:
prometheus-operator allows to specify governing serviceName for our prometheus deployment

this will disable the creation of the `prometheus-operated` service and assign the service created by `kube-prometheus-stack` as the governing service of our statefulset

prometheus-operated service is used by default by thanos query store endpoints. need to remove this dependency before disabling creation of `prometheus-operated` service


viennaa (approved):
Thanks

trouaux:
> Thanks

WIP upstream charts don't allow to pass templated values

richardtief:
@trouaux we upgrade the kube-prometheus-stack with #714

trouaux:
thanos depend on operated prometheus service

PR #723, opened 2025-03-19
feat(alerts) use servicename spec option for alertmanager

trouaux:
specifying serviceName will disable prometheus-operator from creating `alertmanager-operated` service


ibakshay (approved):

richardtief (approved):
Thanks.

Issue #725, opened 2025-03-19
[thanos] avoid missing kube-monitoring sidecar


trouaux:
### Priority

(Low) Something is a little off

### Is your feature request related to a problem?

thanos plugin can get in a READY state even if `kube-monitoring` does not have thanos sidecar deployed

catch this miss configuration when deploying thanos to avoid false positive deployments

### Additional context.

https://github.com/cloudoperators/greenhouse-extensions/tree/main/thanos#object-store-preparation

trouaux:
waiting on https://github.com/cloudoperators/greenhouse-extensions/issues/615

Issue #615, opened 2025-01-31
chore(actions): Incorporate Dependent Plugins in Helm Lint & Helm Test


richardtief:
## Issue Description

Currently, the helm-lint and helm test GitHub workflows do not consider dependent plugins, which may lead to incomplete or failing tests. To improve this, we need a mechanism to define and include dependencies for each plugin.

## Proposed Solution

- Introduce a `dependency.yaml` file in every plugin folder.
- This file will specify whether dependencies exist and, if so, which ones are required.
- Update the GitHub workflow to:
- Parse the dependency.yaml file.
- Install the necessary dependencies before running helm lint and helm test.

## Acceptance Criteria

- Each plugin folder contains an optional dependency.yaml file.
- The GitHub workflow reads this file and installs dependencies accordingly.
- helm lint and helm test execute successfully, considering dependencies.

## Additional Context

This will ensure more robust plugin testing and reduce failures due to missing dependencies.

timojohlo:
As a suggestion: the README of each plugin will get a section `Dependencies` filled by the entries of the `depedency.yaml`

PR #727, opened 2025-03-20
test(thanos) check for thanos sidecar availability e2e

trouaux:
we had a case where a user deployed kube-monitoring without sidecar, then deployed thanos.

thanos plugin status was ready even though there is no sidecar and no data being written to long term storage

adding helm test to check that sidecar container is present

closing
https://github.com/cloudoperators/greenhouse-extensions/issues/725

### DIFF
--- a/thanos/charts/templates/tests/test-thanos-config.yaml
+++ b/thanos/charts/templates/tests/test-thanos-config.yaml
@@ -62,4 +62,11 @@ data:
       verify "there is 1 servicemonitor named '{{ .Release.Name }}'"
     }
     {{ end }}
+
+    {{ if not .Values.thanos.query.standalone }}
+    @test "Verify prometheus sidecar is running" {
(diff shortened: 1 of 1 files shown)


thanos/charts/templates/tests/test-thanos-config.yaml
```diff
@@ -62,4 +62,11 @@ data:
       verify "there is 1 servicemonitor named '{{ .Release.Name }}'"
     }
     {{ end }}
+
+    {{ if and (not .Values.thanos.query.standalone) (not .Values.ci.enabled) }}
```
ibakshay, line 66: You will be able to test this use case once this issue is done https://github.com/cloudoperators/greenhouse-extensions/issues/615 :)

ibakshay (approved):
LGTM! splendid work @trouaux 🚀

PR #732, opened 2025-03-24
fix(kube-monitoring) retention default

trouaux:
removed over write from plugin preset
https://github.com/sapcc/helm-charts/commit/f64f7df2aaa35518092948778bc6816e5d334519

going back to upstream default value


kube-monitoring/plugindefinition.yaml
```diff
@@ -46,8 +46,8 @@ spec:
       type: list
     - name: kubeMonitoring.prometheus.prometheusSpec.retention
       description: How long to retain metrics
-      default: "14d"
-      required: true
+      default: "10d"
+      required: false
```
richardtief, line 50: What's wrong with 14d retention as default?

richardtief (commented):
Thanks. One question to this below.

kube-monitoring/plugindefinition.yaml
richardtief: https://github.com/sapcc/helm-charts/commit/f64f7df2aaa35518092948778bc6816e5d334519#r154229431

trouaux: 10d and 14d could still be too big defaults for the future, was thinking of removing this entry from plugindefinition since 10d is upstream chart default

richardtief (approved):
Thanks.

PR #741, opened 2025-03-25
feat(thanos) query stores update + ruler servicename

trouaux:
updating query store flags to avoid issues for future releases since current flags are being deprecated

\+

ruler serviceName to avoid creation of unused thanos-ruler-operated service and specify correct governing service name


thanos/charts/templates/query/sd-config.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "release.name" . }}-sd-config
```
richardtief, line 4: The ConfigMap is missing the default labels. 
```suggestion
metadata:
  labels:
    {{- include "plugin.labels" . | nindent 4 }}
    {{- include "thanos.labels" . | nindent 4 }}
    {{- if .Values.thanos.query.deploymentLabels }}
    {{ toYaml .Values.thanos.query.deploymentLabels | nindent 4 }}
    {{- end }}
  name: {{ include "release.name" . }}-sd-config
```

thanos/charts/templates/query/deployment.yaml
```diff
@@ -53,18 +53,8 @@ spec:
...
-        {{- fail "WARNING Thanos Query will have not data. Either unset thanos.query.standalone or specify thanos.query.stores." -}}
-        {{- end }}
-        {{- range .Values.thanos.query.stores }}
-        - --store={{ . }}
-        {{- end }}
-        {{- if not .Values.thanos.query.standalone }}
-        - --store=prometheus-operated:10901
-        - --store={{ include "release.name" . }}-store:{{ default "10901" .Values.thanos.grpcAddress }}
-        {{- end }}
+        # --store.sd-files will be deprecated in next release. Correct flag will be --endpoint.sd-config-file
```
richardtief, line 56: Why not using `--endpoint.sd-config-file` straight?

richardtief (changes requested):
Thanks. Two small comments below.

thanos/charts/templates/query/deployment.yaml
trouaux: i tried it and don't think it's implemented yet in current version

richardtief: I can see it in the docs: https://thanos.io/tip/components/query.md/#:~:text=%2D%2Dendpoint.sd%2Dconfig%2Dfile

richardtief: It should work according to this PR: https://github.com/thanos-io/thanos/pull/7890. Could you please create a follow up issue to keep this in mind?

viennaa:
```suggestion
        # --store.sd-files will be deprecated in next release. Correct flag will be --endpoint.sd-config-file (available with v0.38)
```

it is available in v0.38

currently we have [v0.38.0-rc.0](https://github.com/thanos-io/thanos/releases/tag/v0.38.0-rc.0) as a prerelease, as soon as this is in, we can switch.

trouaux: you are looking at the tip
that's our docs https://thanos.io/v0.37/components/query.md/#flags

viennaa (approved):
thanks

richardtief (approved):
Thanks

thanos/charts/templates/query/deployment.yaml
richardtief: Thanks. You are right.

PR #748, opened 2025-03-26
feat(thanos): Expose standalone Query as plutono and perses datasource

richardtief:


thanos/charts/values.yaml
```diff
@@ -152,6 +152,26 @@ thanos:
...
+
+    ## Expose Thanos Query as a Perses datasource
+    ##
+    persesDatasource:
+      # -- Creates a Perses datasource for standalone Thanos Query
+      create: false
+
+      # -- Label selectors for the Perses sidecar to detect this datasource.
+      selector: {}
+        # perses-global: '"true"'
```
ibakshay, line 173: Could you tell, How this selector will work? 
from my understanding, Perses will detect all the configmaps with the labels mentioned in [perses/charts/values.yaml](https://github.com/cloudoperators/greenhouse-extensions/blob/thanos%3Bcreate-plutono-perses-datasource/perses/charts/values.yaml#L155-L157)

thanos/charts/values.yaml
```diff
@@ -152,6 +152,26 @@ thanos:
...
 
+    ## Expose Thanos Query as a Plutono datasource
+    ##
+    plutonoDatasource:
+      # -- Creates a Perses datasource for standalone Thanos Query
+      create: false
+
+      # -- Label selectors for the Plutono sidecar to detect this datasource.
+      selector: {}
+        # plutono-global: '"true"'
```
ibakshay, line 163: Could you tell, How this selector will work? 
from my understanding, Plutono will detect all the configmaps with the labels mentioned in [plutono/charts/values.yaml](https://github.com/cloudoperators/greenhouse-extensions/blob/thanos%3Bcreate-plutono-perses-datasource/plutono/charts/values.yaml#L877-L879

thanos/charts/values.yaml
richardtief: For in-cluster deployments, we keep the default selectors. The idea is to provide a global Perses plugin to retrieve the global resources (data source and dashboards).

thanos/charts/values.yaml
richardtief: Same for Plutono.

trouaux (approved):

thanos/charts/values.yaml
ibakshay: Ah okay! That means, There will be a separate global Perses/Plutono Plugin living in the central cluster with global thanos as datasource?

PR #756, opened 2025-03-28
chore(alerts): remove lookup function to resolve reconciliation loops

richardtief:


viennaa (commented):
Would this end up in a recon loop if no root-ca is present?

richardtief:
No, the condition at the top of the manifest will prevent this. The lookup function seems to produce a diff which led to the reconciliation loops.

viennaa (approved):
Ok. This actually was my question:

```
{{- $intCASecret := (lookup "v1" "Secret" $.Release.Namespace (printf "tls-alerts-%s-root-ca" .Release.Namespace )).data }}
```


root ca gone, reconciliation loop. this might pair with lookup.

PR #772, opened 2025-04-07
feat(opensearch): add ISM policies and index templates for logs data stream

joluc:
## Pull Request Details

This PR adds Index State Management policies and index templates for the logs data stream. 

- Add ISM policy (`logs-rollover-policy`) that:
  - Rolls over indices after 7 days, 30GB size, or 50M documents
  - Automatically deletes indices after 30 days
  - Targets all indices matching the pattern `logs*`
- An index template (`logs-index-template`) that:
  - Configures default settings for all log indices (1 shard, 1 replica)
  - Sets up proper mappings for timestamps and message fields
  - Defines a data stream 
  
Due to an upstream missing field in the helm chart, we need to duplicate the `OpensearchIndexTemplate` resource. I am going to contribute this fix upstream at [issue #990](https://github.com/opensearch-project/opensearch-k8s-operator/issues/990).

## Breaking Changes

No

## Issues Fixed

X

## Other Relevant Information

X


timojohlo (approved):
🚀

PR #779, opened 2025-04-09
fix(logs): parsing for kvm and ceph

timojohlo:
## Pull Request Details

fixes parsing for ceph and kvm and sets `imagePullPolicy: Always` for the Collector image to update properly as we are pulling from `:main` tag.

### DIFF
--- a/logs/charts/Chart.yaml
+++ b/logs/charts/Chart.yaml
@@ -4,7 +4,7 @@
 apiVersion: v2
 appVersion: v0.121.0
 name: logs-operator
-version: 0.8.5
+version: 0.8.6
 description: OpenTelemetry Operator Helm chart for Kubernetes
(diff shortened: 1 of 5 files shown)


logs/charts/values.yaml
```diff
@@ -30,7 +30,7 @@ logs-operator:
     # -- overrides the default image repository for the OpenTelemetry Collector image.
       repository: ghcr.io/cloudoperators/opentelemetry-collector-contrib
     # -- overrides the default image tag for the OpenTelemetry Collector image.
-      tag: main
+      tag: 1e76287
```
ibakshay, line 33: Are you pointing to a particular commit?

timojohlo: I am pointing to a dedicated docker image version [here](https://github.com/cloudoperators/greenhouse-extensions/pkgs/container/opentelemetry-collector-contrib/392648082?tag=1e76287)
This is to make sure the collectors will be update properly when updating the image and also mitigating relying on `ImagePullPolicy: Always`

PR #794, opened 2025-04-16
fix(opensearch): Change label and scope for opensearch crds

joluc:
## Pull Request Details

This PR updates the OpenSearch CRDs to:
- Use the standard opensearch namespace label instead of the demo-specific demo label
- Change the scope from `Namespaced` to `Cluster` to support use in multiple namespaces

## Breaking Changes

None. 

## Issues Fixed

Currently the plugin cannot be deployed because demo namespace is protected.


richardtief (approved):
😉

PR #805, opened 2025-04-23
fix(logs): updates service/telemetry due to upstream API change

olandr:
Bumps opentelemetry-collector-contrib and opentelemetry-operator versions. Also resolves a breaking change as described here:
- https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-metrics
- https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.123.0


logs/charts/templates/logs-collector.yaml
```diff
@@ -332,8 +332,13 @@ spec:
 {{- if .Values.openTelemetry.prometheus.podMonitor.enabled }}
       telemetry:
         metrics:
-          address: 127.0.0.1:8888
           level: detailed
+          readers:
+          - pull:
+              exporter:
+                prometheus:
+                  host: 127.0.0.1
```
trouaux, line 340: should we add quotes like in the docs ?

logs/charts/Chart.yaml
```diff
@@ -2,9 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 apiVersion: v2
-appVersion: v0.121.0
+appVersion: v0.122.0
```
trouaux, line 5: is this used ?

olandr: Not sure, it was added here #246 and here I am bumping it to stay consistent.

Kuckkuck (approved):

logs/charts/Chart.yaml
olandr: Decided to put it as the v0.124.1 which is consistent with the docker image used, see https://github.com/cloudoperators/greenhouse-extensions/commit/469df1a305ba96f8290b37766eefe343fdc589d9

trouaux (approved):

PR #809, opened 2025-04-24
chore(alerts): restore webhook secret and enhance Slack notification template 

richardtief:
closing #1006


alerts/charts/Chart.yaml
```diff
@@ -8,7 +8,7 @@ maintainers:
 name: alerts
 sources:
   - https://github.com/cloudoperators/greenhouse-extensions
-version: 0.23.2
+version: 0.23.1
```
viennaa, line 11: is this a change without a change?

alerts/charts/templates/am-base-config.yaml
```diff
@@ -58,4 +58,23 @@ spec:
             name: {{ $.Release.Name }}-webhook-urls
```
viennaa, line 58: how was this secret provided earlier, before it has been attached to the config here?

richardtief: It was remove accidentally.

PR #828, opened 2025-04-28
fix(thanos) integration test mount tls

trouaux:


thanos/charts/templates/tests/test-thanos-config.yaml
```diff
@@ -28,7 +28,7 @@ data:
       for endpoint in $endpoints; do
         if [[ $endpoint =~ :443$ ]];
         then
-          run grpc_health_probe -addr=${endpoint} -tls -tls-ca-cert /tls-assets/ca.crt
+          run grpc_health_probe -addr=${endpoint} -tls -tls-no-verify
```
ibakshay, line 31: why this change required?

trouaux: `# failed to initialize tls credentials. error=failed to load root CA certificates from file (/tls-assets/ca.crt) error=open /tls-assets/ca.crt: no such file or directory`

cert I use below is not available on all thanos deployments, global does not create a GRPC ingress
https://github.com/cloudoperators/greenhouse-extensions/blob/main/thanos/charts/templates/tests/test-thanos-config.yaml#L89

so this mount is not happening
https://github.com/cloudoperators/greenhouse-extensions/blob/main/thanos/charts/templates/tests/test-thanos.yaml#L35-L40

we could specify random name and cert-manager would handle it but don't think it's relevant for integration tests

i'm actually considering removing this tls all together from the test, any objection to use tls no verify for health checks ? @richardtief @viennaa

richardtief: Are we testing within the cluster? Then we don't need a certificate for the test, right?

trouaux: global query endpoint config is full of external endpoints

ibakshay (approved):
LGTM!

PR #829, opened 2025-04-29
feat(dockerfiles) docker images versionning

trouaux:
1. extracting version from LABEL of Dockerfile
2. check if version is already present in image repository
3. stop workflow if bump is required


kengou (commented):
Versioning is done via tagging, if you push a tag a new image will be created via that tag.
I don't see the benefit of manually maintaining it inside the docker image
https://github.com/cloudoperators/greenhouse-extensions/blob/7e75524651527ca0a555d6c4ed642d89bbedf5e9/.github/workflows/docker-build.yaml#L5-L11

trouaux:
> Versioning is done via tagging, if you push a tag a new image will be created via that tag. I don't see the benefit of manually maintaining it inside the docker image
> 
> https://github.com/cloudoperators/greenhouse-extensions/blob/7e75524651527ca0a555d6c4ed642d89bbedf5e9/.github/workflows/docker-build.yaml#L5-L11

got it thanks for the review, we will start using tags

PR #832, opened 2025-04-29
fix(thanos) integration-test-versionning

trouaux:
- versioned tag for integration test
- clean up unused values


ibakshay (approved):
LGTM!

PR #834, opened 2025-04-30
feat(repo) make pint

trouaux:
- added renovate config for pint and helm versions
- not deleting binaries if version is matching config (helm-docs, pint, helm)
- make pint PLUGIN=kube-monitoring

Screenshot 2025-04-30 at 12 54 17 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ YQ ?= $(LOCALBIN)/yq
 
 KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
 YQ_INSTALL_SCRIPT ?= https://github.com/mikefarah/yq/releases/latest/download/yq_$(OS)_$(ARCH)
-HELM_DOCS_REPO ?= github.com/norwoodj/helm-docs/cmd/helm-docs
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -15,6 +15,8 @@ endif
 KUSTOMIZE_VERSION ?= 5.6.0
 YQ_VERSION ?= v4.45.1
 HELM_DOCS_VERSION ?= 1.14.2
+PINT_VERSION ?= 0.73.5
+HELM_VERSION ?= 3.17.3
```
kengou, line 19: If you're adding new tools here, please add them to [renovate.json](https://github.com/cloudoperators/greenhouse-extensions/blob/ed375bb797d452ad6abe01244db65b9aba3fbbe6/.github/renovate.json) as well.

here is a reference: https://github.com/cloudoperators/greenhouse-extensions/blob/ed375bb797d452ad6abe01244db65b9aba3fbbe6/.github/renovate.json#L98-L110

trouaux: thanks for the review, I added the config.

had a question regarding the 'Removing it before installing.' condition when installing binaries.
does it make sense to not delete if present ? and maybe have a make command to delete bin folder ?

kengou: I would not delete it, of course it could lead to problems when you change the version of the tool

trouaux: thanks, bandwidth saving feels worth it

trouaux:
@kengou I addressed what we discussed, can you have a second look

kengou (approved):

IvoGoman (approved):

ibakshay:
@trouaux does it make sense to integrate pint in the GitHub Actions PR workflow? same like `helm-docs`?

PR #835, opened 2025-04-30
feat(opensearch): add labels for ServiceMonitor

joluc:
## Pull Request Details

Allow to set custom labels for the ServiceMonitor to expose OpenSearch metrics to Prometheus.

## Breaking Changes

None

## Issues Fixed

N/A

### DIFF
--- a/opensearch/chart/values.yaml
+++ b/opensearch/chart/values.yaml
@@ -208,6 +208,9 @@ cluster:
         tlsConfig:
           insecureSkipVerify: true
 
+        # -- ServiceMonitor labels
+        labels: {}
+
       # -- List of Opensearch plugins to install
       pluginsList: []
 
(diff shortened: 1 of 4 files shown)


opensearch/chart/values.yaml
```diff
@@ -208,6 +208,9 @@ cluster:
         tlsConfig:
           insecureSkipVerify: true
 
+        # -- ServiceMonitor labels
+        labels: {}
```
trouaux, line 212: can't find this value existing in cluster helm charts

joluc: It's indeed missing in the helm chart upstream. They most probably just missed that. Will open a PR. The whole block `monitoring` get's applied in the OpenSearchCluster resource under `general` [here](https://github.com/opensearch-project/opensearch-k8s-operator/blob/14a8251/charts/opensearch-cluster/templates/cluster.yaml#L25). The labels field is defined in the [CRD](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/charts/opensearch-operator/files/opensearch.opster.io_opensearchclusters.yaml#L3409-L3412). The operator then loads the labels and creates a ServiceMonitor resource [here](https://github.com/opensearch-project/opensearch-k8s-operator/blob/b80b606964dccb0dd18d1ea01d0c0815eaf0be68/opensearch-operator/pkg/builders/cluster.go#L1226-L1228).

trouaux: got it thanks for the details

trouaux (approved):

ibakshay (approved):

PR #847, opened 2025-05-07
fix(makefile) mkdir bin if not present

trouaux:
missing /bin check for new binaries


Makefile
```diff
@@ -82,7 +87,8 @@ pint-install:
 
 ## Download `pint` locally if necessary
```
IvoGoman, line 88:
```suggestion
## Download `helm` locally if necessary
```

IvoGoman (approved):

PR #851, opened 2025-05-09
fix(.github): updates url for HELM_DOCS_REPO to resolve 404

olandr:
The HELM_DOCS_REPO install url yields a 404 on the workflow run https://github.com/cloudoperators/greenhouse-extensions/actions/runs/14927089409/job/41934304611?pr=766

Seemingly from the list of releases we need to use x86_64 instead, see https://github.com/norwoodj/helm-docs/releases/tag/v1.14.2

---------

Signed off by: Simon Olander (simon.olander@sap.com)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 OS := $(shell uname -s | tr '[:upper:]' '[:lower:]')
 
 # Detect ARCH (AMD64 or ARM64)
+UNAME_M := $(shell uname -m)
 UNAME_P := $(shell uname -p)
 ARCH :=
 ifeq ($(UNAME_P),x86_64)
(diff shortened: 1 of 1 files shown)


ibakshay (approved):
🚀

ibakshay:
@olandr it looks like this approach doesn't work when you execute the `make generate-readme PLUGIN=PLUGIN_NAME` from ARM based Macbook. Could you please check?

ibakshay:
https://github.com/cloudoperators/greenhouse-extensions/blob/009be4400d71a9c0fb1bb0545b4f4750afecbc74/Makefile#L5
```
UNAME_P := $(shell uname -m) should work. 
```

olandr:
> https://github.com/cloudoperators/greenhouse-extensions/blob/009be4400d71a9c0fb1bb0545b4f4750afecbc74/Makefile#L5
> 
> Below code snippet should work. `-m` instead of `-p`
> 
> ```
> UNAME_P := $(shell uname -m)
> ```

Yeah seems like a [better idea](https://www.gnu.org/software/coreutils/manual/html_node/uname-invocation.html#index-_002dp-12).

ibakshay (approved):
LGTM!

PR #854, opened 2025-05-09
feat(opensearch): support cert-manager certs and disable self-signed

joluc:
## Pull Request Details

This change:
- Disables the operators automatic generation of self-signed certificates
- Creates three new certificates:
  - `opensearch-http-cert` for the REST API
  - `opensearch-transport-cert` for node-to-node communication
  - `opensearch-admin-cert` for admin operations

## Breaking Changes

N/A

## Issues Fixed

N/A

### DIFF
--- a/opensearch/README.md
+++ b/opensearch/README.md
@@ -90,8 +90,8 @@ This guide provides a quick and straightforward way to use **OpenSearch** as a G
 | cluster.cluster.dashboards.service.type | string | `"ClusterIP"` | dashboards service type |
 | cluster.cluster.dashboards.tls.caSecret | object | `{}` | Secret that contains the ca certificate as ca.crt. If this and generate=true is set the existing CA cert from that secret is used to generate the node certs. In this case must contain ca.crt and ca.key fields |
 | cluster.cluster.dashboards.tls.enable | bool | `false` | Enable HTTPS for dashboards |
-| cluster.cluster.dashboards.tls.generate | bool | `true` | generate certificate, if false secret must be provided |
-| cluster.cluster.dashboards.tls.secret | string | `nil` | Optional, name of a TLS secret that contains ca.crt, tls.key and tls.crt data. If ca.crt is in a different secret provide it via the caSecret field |
+| cluster.cluster.dashboards.tls.generate | bool | `false` | generate certificate, if false secret must be provided |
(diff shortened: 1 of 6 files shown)


opensearch/README.md
```diff
@@ -90,8 +90,8 @@ This guide provides a quick and straightforward way to use **OpenSearch** as a G
 | cluster.cluster.dashboards.service.type | string | `"ClusterIP"` | dashboards service type |
 | cluster.cluster.dashboards.tls.caSecret | object | `{}` | Secret that contains the ca certificate as ca.crt. If this and generate=true is set the existing CA cert from that secret is used to generate the node certs. In this case must contain ca.crt and ca.key fields |
 | cluster.cluster.dashboards.tls.enable | bool | `false` | Enable HTTPS for dashboards |
-| cluster.cluster.dashboards.tls.generate | bool | `true` | generate certificate, if false secret must be provided |
-| cluster.cluster.dashboards.tls.secret | string | `nil` | Optional, name of a TLS secret that contains ca.crt, tls.key and tls.crt data. If ca.crt is in a different secret provide it via the caSecret field |
+| cluster.cluster.dashboards.tls.generate | bool | `false` | generate certificate, if false secret must be provided |
+| cluster.cluster.dashboards.tls.secret | object | `{"name":"opensearch-http-cert"}` | Optional, name of a TLS secret that contains ca.crt, tls.key and tls.crt data. If ca.crt is in a different secret provide it via the caSecret field |
```
olandr, line 94: Why is this an object?

olandr, line 94: I am not understanding the if-stmt here:
> If ca.crt is in a different secret provide it via the caSecret field 

why is this only relevant for `ca.crt`? Where is this _field_ that I can set?

olandr (changes requested):
Nice stuff, just two questions on the docs here.

opensearch/README.md
joluc: The CRD actually defines it as an object (see lines 2821–2832 in the [OpenSearchCluster schema](https://github.com/opensearch-project/opensearch-k8s-operator/blob/main/opensearch-operator/config/crd/bases/opensearch.opster.io_opensearchclusters.yaml#L2821-L2832)). In the upstream values.yaml, that object was commented out without empty braces, so the generator mistakenly treated it as a string which isn’t correct. :)

joluc: When requesting a certificate via cert-manager, the ca.crt field is automatically included. If I want to use my own CA, I can provide it manually by referencing a separate secret. In our case, cert-manager adds the DigiCert CA to the certificate secret.

olandr (approved):
Alright, I think the README is not trivial to understand. But that might be a problem upstream instead. LGTM

PR #858, opened 2025-05-12
Add self-monitoring Perses dashboards for Perses

ibakshay:


richardtief (approved):
Thanks for sorting this out. 🙂

PR #886, opened 2025-05-20
feat(repo-guard): move from github-guard, add resources, update plugin definition

onuryilmaz:
## Pull Request Details
- Rename `github-guard` plugin to `repo-guard`
- Add resources to be managed by the plugin
- Update documentation
- Update plugin definition with option values


repo-guard/charts/templates/distributionlistapi.yaml
```diff
@@ -0,0 +1,18 @@
+{{ if .Values.distributionListAPI }}
```
uwe-mayer, line 4: We are only expecting single DLs per chart? No `range` needed?

repo-guard/charts/values.yaml
```diff
@@ -1,23 +1,81 @@
...
-        - ALL
-    image:
+manager:
+  containerSecurityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop:
+      - ALL
+  image:
       repository: keppel.global.cloud.sap/ccloud/github-guard-greenhouse
```
uwe-mayer, line 11: We cannot have internal registries/images in `greenhouse-extensions`, pls publish to `ghcr.io`

repo-guard/charts/templates/deployment.yaml
```diff
@@ -52,9 +50,9 @@ spec:
             port: 8082
           initialDelaySeconds: 5
           periodSeconds: 10
-        resources: {{- toYaml .Values.controllerManager.manager.resources | nindent 10
+        resources: {{- toYaml .Values.manager.resources | nindent 10
           }}
-        securityContext: {{- toYaml .Values.controllerManager.manager.containerSecurityContext
+        securityContext: {{- toYaml .Values.manager.containerSecurityContext
```
uwe-mayer, line 55: Is there scenarios where these need to be escalated beyond the defaults in `values.yaml`?

repo-guard/charts/templates/github.yaml
```diff
@@ -0,0 +1,24 @@
+{{- range $githubKey, $githubItem := .Values.githubs }}
+apiVersion: githubguard.sap/v1
```
uwe-mayer, line 5: Comment, no blocker:
Refactor to use `repoguard.sap/v1`

repo-guard/charts/templates/distributionlistapi.yaml
onuryilmaz: DL API will be replaced with a generic custom resource in the next releases, then we can refactor to have multiple instances

repo-guard/charts/templates/deployment.yaml
onuryilmaz: no, since repo-guard will be singleton in the cluster these values could be managed by cluster operators

repo-guard/charts/templates/github.yaml
onuryilmaz: https://github.com/cloudoperators/repo-guard/issues/14

repo-guard/charts/templates/deployment.yaml
```diff
@@ -52,9 +52,5 @@ spec:
           periodSeconds: 10
         resources: {{- toYaml .Values.manager.resources | nindent 10
           }}
-        securityContext: {{- toYaml .Values.manager.containerSecurityContext
-          | nindent 10 }}
-      securityContext:
-        runAsNonRoot: true
```
abhijith-darshan, line 60: Why are we removing the security context?

PR #895, opened 2025-05-21
feat(renovate) group makefile

trouaux:


.github/renovate.json
```diff
@@ -267,7 +267,12 @@
...
       "groupName": "integration-test Dockerfile dependencies"
+    },
+    {
+      "matchManagers": ["regex"],
+      "matchPaths": ["Makefile"],
+      "groupName": "Makefile dependencies"
     }
   ],
-  "separateMinorPatch": true
+  "separateMinorPatch": false
```
trouaux, line 277: @kengou do you remember why we set this to true ?

kengou: If set to `true`, Renovate will separate `minor` and `patch` updates into separate branches.

PR #900, opened 2025-05-23
feat(automation) add workflow_dispatch

trouaux:


.github/workflows/helm-lint-test.yaml
```diff
@@ -2,6 +2,7 @@ name: "Helm lint and tests"
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+  workflow_dispatch:
```
IvoGoman, line 5: I think this YAML is invalid. Should be either without the `:` or with the empty object notation
```suggestion
  workflow_dispatch: {}
```

PR #903, opened 2025-05-23
feat(automation) helm-test pod logs on failure

trouaux:
see this workflow run as reference

https://github.com/cloudoperators/greenhouse-extensions/actions/runs/15215784964/job/42800864630

### DIFF
--- a/.github/workflows/helm-lint-test.yaml
+++ b/.github/workflows/helm-lint-test.yaml
@@ -75,4 +75,15 @@ jobs:
 
       - name: Run chart integration tests
         if: steps.chart-list-changed.outputs.changed == 'true'
-        run: ct lint-and-install --config .github/configs/helm-lint-test.yaml
+        run: ct lint-and-install --config .github/configs/helm-lint-test.yaml --skip-clean-up
+      
+      - name: Print pod logs on failure
+        if: failure() && steps.chart-list-changed.outputs.changed == 'true'
+        run: |
+          echo "🔍 Printing logs for failed pods..."
+          for ns in $(kubectl get ns -o jsonpath='{.items[*].metadata.name}'); do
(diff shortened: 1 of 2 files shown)


.github/workflows/helm-lint-test.yaml
```diff
@@ -75,4 +75,15 @@ jobs:
...
       - name: Run chart integration tests
         if: steps.chart-list-changed.outputs.changed == 'true'
-        run: ct lint-and-install --config .github/configs/helm-lint-test.yaml
+        run: ct lint-and-install --config .github/configs/helm-lint-test.yaml --skip-clean-up
+      
+      - name: Print pod logs on failure
+        if: failure() && steps.chart-list-changed.outputs.changed == 'true'
+        run: |
+          echo "🔍 Printing logs for failed pods..."
+          for ns in $(kubectl get ns -o jsonpath='{.items[*].metadata.name}'); do
```
ibakshay, line 84: We don't have to iterate over all the namespaces. The namespace will be always `default`. We have declared the namespace here -https://github.com/cloudoperators/greenhouse-extensions/blob/ci-test-pod-logs/.github/configs/helm-lint-test.yaml#L9

ibakshay (changes requested):

.github/workflows/helm-lint-test.yaml
```diff
@@ -75,4 +75,15 @@ jobs:
 
       - name: Run chart integration tests
         if: steps.chart-list-changed.outputs.changed == 'true'
-        run: ct lint-and-install --config .github/configs/helm-lint-test.yaml
+        run: ct lint-and-install --config .github/configs/helm-lint-test.yaml --skip-clean-up
```
ibakshay, line 78: Suggestion: Please include this flag `skip-clean-up` directly inside the `.github/configs/helm-lint-test.yaml` instead of hard-coding here. 
```suggestion
        run: ct lint-and-install --config .github/configs/helm-lint-test.yaml
```

PR #911, opened 2025-05-26
feat(opensearch): allow configuring HTTP DNS names via plugin

joluc:
## Pull Request Details

This PR includes the following changes:

- Refactored certificate Helm charts: moved more configuration into `values.yaml`
- Split issuer and certificate definitions into separate files
- Made the HTTP DNS name configurable via the plugin interface

### DIFF
--- a/opensearch/chart/values.yaml
+++ b/opensearch/chart/values.yaml
@@ -1,18 +1,50 @@
 # SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse contributors
 # SPDX-License-Identifier: Apache-2.0
 
-global:
-  ## common greenhouse values to use
-  ##
-  greenhouse:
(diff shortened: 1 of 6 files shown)


opensearch/chart/values.yaml
```diff
@@ -1,19 +1,51 @@
...
-  ##
-  greenhouse:
-    # base domain of the greenhouse organisation
-    baseDomain:
-  certManager:
-    enable: true
-    issuerRef:
+# -- Enable cert-manager integration for issuing TLS certificates
+certManager:
+  enable: true
```
trouaux, line 6: this will make the README more readable

```suggestion

certManager:
  # -- Enable cert-manager integration for issuing TLS certificates
  enable: true
```

joluc: Thanks, good point! :)

PR #921, opened 2025-05-28
feat(opensearch): add configurable user credentials

joluc:
## Pull Request Details

This PR improves the OpenSearch cluster configuration by adding support for configurable user credentials with failover capabilities.

- Configurable user credentials: New `usersCredentials` section in values.yaml allowing override of usernames and passwords for OpenSearch users from Plugin
- Restructured user configuration to separate credentials from user role definitions
- Failover user support: Secondary user (`logs2`) configuration for redundancy

### DIFF
--- a/opensearch/chart/templates/users-secret.yaml
+++ b/opensearch/chart/templates/users-secret.yaml
@@ -2,15 +2,15 @@
 # SPDX-License-Identifier: Apache-2.0
 
 {{- $labels := include "opensearch-cluster-local.labels" . }}
-{{- range .Values.cluster.users }}
+{{- range $username, $credentials := .Values.cluster.usersCredentials }}
 ---
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ .name }}-credentials
+  name: {{ $username }}-credentials
   labels: {{ $labels | nindent 4 }}
 type: Opaque
 data:
-  username: {{ .name | b64enc }}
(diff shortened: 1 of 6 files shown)


opensearch/chart/templates/users-secret.yaml
```diff
@@ -2,15 +2,20 @@
...
-  name: {{ .name }}-credentials
+  name: {{ $username }}-credentials
   labels: {{ $labels | nindent 4 }}
 type: Opaque
 data:
-  username: {{ .name | b64enc }}
-  password: {{ default (printf "%s%s" (randAlphaNum 15) (index (list "!" "@" "#" "$" "%" "^" "&" "*") (randInt 0 8))) .password | b64enc }}
+  username: {{ default $username $credentials.username | b64enc }}
+  password: {{ default (index $defaultPasswords $username) $credentials.password | b64enc }}
 {{- end }}
```
olandr, line 21: I would suggest changing the key here to something other than `$username`. It is a bit strange to (on line 19) have this logic:

```
username: {{ default $username $credentials.username | b64enc }}
```
I think this is a bit confusing: `$credentials.username` is for the key username instead of the key `$username`:

Similar on line 20:
```
password: {{ default (index $defaultPasswords $username) $credentials.password | b64enc }}
```

Maybe you could call it `failoverUser`, `failoverObject` or something similar?

olandr (changes requested):
Great stuff, I have one small comment on the choice of a variable name.

olandr (approved):
Looks more clearer now!

PR #922, opened 2025-05-28
fix(thanos) update metrics names

trouaux:
fixing alerts from absent-metrics-operator

### DIFF
--- a/thanos/charts/Chart.yaml
+++ b/thanos/charts/Chart.yaml
@@ -11,7 +11,7 @@ maintainers:
 name: thanos
 sources:
   - https://github.com/cloudoperators/greenhouse-extensions
-version: 0.5.23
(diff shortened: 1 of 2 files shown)


thanos/charts/Chart.yaml
```diff
@@ -11,7 +11,7 @@ maintainers:
 name: thanos
 sources:
   - https://github.com/cloudoperators/greenhouse-extensions
-version: 0.5.21
+version: 0.5.23
```
ibakshay, line 14: Why not 0.5.22? 
```suggestion
version: 0.5.22
```

trouaux: correct, thanks

ibakshay (approved):

PR #975, opened 2025-06-18
fix(opensearch): dual certificates for http endpoint

joluc:
What changed

- Dual certificates for HTTP:
  - Internal cert (self-signed CA) kept for operator-to-cluster calls that use hard-coded service DNS names.
  - External cert (opensearch-http-cert-external) used by the Ingress for log ingestion and TLS challenges.
- Bundled security-config Secret: default config.yml, users, roles; admin password

Internal calls need a cert that covers *.svc.cluster.local, while the public endpoint needs a browser-trusted cert for the FQDN. The dual-cert flow solves both without manual patching.


opensearch/plugindefinition.yaml
```diff
@@ -53,18 +53,47 @@ spec:
...
       required: true
       type: secret
     - name: cluster.usersCredentials.logs2.username
-      description: "Secondary OpenSearch user username for failover authentication"
+      description: "Secondary OpenSearch username for failover log ingestion."
       required: true
       type: secret
     - name: cluster.usersCredentials.logs2.password
-      description: "Secondary OpenSearch user password for failover authentication"
+      description: "Secondary OpenSearch user password for failover log ingestion."
```
olandr, line 68: I disagree with these changes, imo "failover for log ingestion" does not make a lot of sense. Failover _is_ authentication related no? Sure the user is technically for log ingestion (so those changes are fine), but not the failover mechanism.

If anything (not saying you should use this because it is quite verbose) it should be closer to this:

```
Secondary user, for log ingestion, password to support failover.
```

Same for username.

But I would lean on not change it at all.

olandr (changes requested):
Great stuff. One small comment on the docs.

opensearch/chart/values.yaml
```diff
@@ -649,16 +676,22 @@ cluster:
...
-  #      - "admin-role"
+        - "logs-write-role"
+    - name: "logs-read"
+      backendRoles:
+        - "CC_IAS_OPERATIONS_UI_KIBANA_SUPPORT"
+      roles:
+        - "logs-read-role"
+    - name: "admin"
+      backendRoles:
+        - "CC_IAS_TEAM_SUPERVISION"
```
timojohlo, line 692: These roles are very specific to our use case imo. I would rather see them as secrets that you can point to in the values.yaml file, what do you think?

timojohlo (changes requested):
Cool to see all these changes coming in! 👍 
One small remark.

opensearch/plugindefinition.yaml
joluc: Yes, let me change that :)

PR #976, opened 2025-06-18
fix(kube-monitoring) blackbox config rename

trouaux:
renaming config.yaml to blackbox.yaml as used by default in blackbox helm charts
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-blackbox-exporter/templates/_helpers.tpl#L183C62-L183C70


kube-monitoring/charts/templates/_helper.tpl
```diff
@@ -129,7 +129,6 @@ release: {{ $root.Release.Name | quote }}
 {{- define "kubeMonitoring.dashboardSelectorLabels" }}
 {{- $path := index . 0 -}}
 {{- $root := index . 1 -}}
-plugin: {{ $root.Release.Name }}
```
viennaa, line 132: why do we need to remove this?

viennaa (commented):
Thanks for the PR. Small thing for my understanding to clarify.

kube-monitoring/charts/templates/_helper.tpl
viennaa: Ok, now that I look at the labels, it looks like we have the same helper twice:

* [kube-prometheus-stack.labels](https://github.com/cloudoperators/greenhouse-extensions/blob/main/kube-monitoring/charts/templates/_helper.tpl#L30-L39) 
* [kubeMonitoring.labels](https://github.com/cloudoperators/greenhouse-extensions/blob/main/kube-monitoring/charts/templates/_helper.tpl#L116-L127)

Do you comply? I think we should split this in 2 PRs. keep this one about the blackbox config change and open a new one for the duplicate labels, where we can remove the helper.

trouaux: agreed

PR #985, opened 2025-06-20
fix(logs) add plugin label to logs alerts

trouaux:
for proper integration with absent metrics operator we need a common label on PrometheusRule CR for the operator to select from

adding `plugin` label to align with other plugins

Screenshot 2025-06-20 at 11 42 51 AM


trouaux:
done in https://github.com/cloudoperators/greenhouse-extensions/pull/987/files

PR #990, opened 2025-06-23
fix(kube-monitoring) absent metrics operator custom rule name

trouaux:
setting `--prom-rule-name` to create absent rules based on plugin name

default being
`{{ if index .metadata.labels "thanos-ruler" }}{{ index .metadata.labels "thanos-ruler" }}{{ else }}{{ index .metadata.labels "prometheus" }}{{ end }}
`


viennaa:
Is this altering the exposed rule name? Can you give an example how it is named after this change?

PR #1003, opened 2025-06-26
feat(opensearch): refactor dashboards user 

joluc:
## Pull Request Details

- Creating a dedicated dashboards user (and optional failover dashboards2 user) for Dashboards access.
- Defining a minimal dashboards-role based on the official kibana_server role, granting only the permissions required for Dashboards to function.


opensearch/chart/values.yaml
```diff
@@ -354,8 +354,8 @@ cluster:
 
       # -- Secret that contains fields username and password for dashboards to use to login to opensearch, must only be supplied
       # if a custom securityconfig is provided
-      opensearchCredentialsSecret: {}
-        # name: "secret-name"
+      opensearchCredentialsSecret:
+        name: "dashboards-credentials"
```
timojohlo, line 358: Is it possible to provide more than one secret reference here? 
Otherwise you will need to manually swap in case of a credential rotation:
`name: "dashboards-credentials"`
with 
`name: "dashboards2-credentials"`?

timojohlo (commented):
LGTM, just a small question.

trouaux:
Screenshot 2025-06-26 at 2 26 47 PM

PR #1024, opened 2025-07-04
fix(logs): duplicate labels from resource/attributes, cleanup alerts

timojohlo:
## Duplicate Labels
In a previous version, the `k8s.namespace.name` and `k8s.node.name` were shipped twice. This change consolidates the labels, so the container logs and k8sevents will ship the same labels (`resource.k8s.namespace.name`, `resource.k8s.node.name`).

## Cleanup Alerts
- Removed not needed alerts
- Disabled operator alerts `ReconcileErrors` and `WorkqueueDepth` 
- increased the rate slightly for `otelcol_receiver_refused_metric_points_total`

### DIFF
--- a/logs/charts/values.yaml
+++ b/logs/charts/values.yaml
@@ -101,6 +101,7 @@ openTelemetry:
       disabled:
         - ReconcileErrors
         - WorkqueueDepth
+        - ReceiverRefusedMetric
 
       # -- Labels for PrometheusRules.
       labels: {}


trouaux:
thanks, i see you found absent metrics alerts

can't find how `ReceiverRefusedMetric` is being disabled ? with that metric `otelcol_receiver_refused_metric_points_total`

PR #1031, opened 2025-07-09
feat(thanos) custom objectstore secret name

trouaux:
allow to pass custom secret name with `.Values.thanos.compactor.objectstoreSecretName`

instead of default `{ include "release.name" . }}-metrics-objectstore`


thanos/charts/values.yaml
```diff
@@ -238,6 +240,9 @@ thanos:
     # -- Adding additional arguments to Thanos Compactor
     additionalArgs: []
 
+    # -- Thanos Compactor ObjestStore secret name
+    objectstoreSecretName:
+
```
richardtief, line 245: This can be omitted in favor of the one in line 29

thanos/charts/values.yaml
```diff
@@ -26,6 +26,8 @@ thanos:
   # @default -- 0.0.0.0:10902
   httpAddress:
 
+  objectstoreSecretName:
+
```
richardtief, line 30: The secret name is used for both the compactor and the store, so this value should be valid for both.

richardtief (changes requested):
Thanks. One nit below.

trouaux:
> Thanks. One nit below.

thanks did some changes. renamed to `existingObjectStoreSecretName`

thanos/charts/values.yaml
```diff
@@ -26,6 +26,9 @@ thanos:
   # @default -- 0.0.0.0:10902
   httpAddress:
 
+  # -- Thanos Compactor ObjestStore secret name
```
richardtief, line 29:
```suggestion
  # -- Use existing objectStorageConfig Secret data and configure it to be used by Thanos Compactor and Store
  # -- https://thanos.io/tip/thanos/storage.md/#s3
```

richardtief (commented):
Not only compactor.

richardtief (approved):

sjentzschsap:
Personally, I would propose
```
secretName: {{ tpl .Values.thanos.existingObjectStoreSecretName . }}
```
instead of
```
secretName: {{ .Values.thanos.existingObjectStoreSecretName | default (include "release.name" . | printf "%s-metrics-objectstore") }}
```

And then in the default values:
```
existingObjectStoreSecretName: '{{ include "release.name" . }}-metrics-objectstore'
```
instead of
```
existingObjectStoreSecretName:
```

Makes it cleaner, simpler, and your default is only defined once where it belongs (in the values.yaml).

But, both would work and solve the issue.

PR #1034, opened 2025-07-09
kube-monitoring: set the plutono datasource as default

richardtief:
Dashboards are missing a pre-selected data source.


richardtief:
Sure! The condition can be added. `{{ if kubeMonitoring.prometheus.prometheusSpec.thanos }}` then thanos datasource is the default.

trouaux (approved):

PR #1039, opened 2025-07-14
fix(opensearch): cleanup roles and certmanager specifics

joluc:
## Pull Request Details

- Move SAP specifics out of the helm chart


timojohlo (approved):
much cleaner, LGTM!

PR #1046, opened 2025-07-21
feat(opensearch): add dashboards external service resource for service proxy

joluc:
## Pull Request Details

The OpenSearch Operator creates the dashboards service dynamically, which means it is not part of the Helm-managed resources. As a result, the service proxy does not recognize it properly, since it only supports services defined within a Helm release.

This PR re-adds an external dashboards service definition to the Helm chart to ensure the service is tracked by Helm and compatible with the proxy requirements.

### DIFF
--- a/opensearch/chart/Chart.yaml
+++ b/opensearch/chart/Chart.yaml
@@ -3,7 +3,7 @@
 
 apiVersion: v2
 name: opensearch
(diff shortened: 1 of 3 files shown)


timojohlo (approved):
LGTM!

PR #1047, opened 2025-07-22
feat(logs): adds v2beta1 version to the otelcol CRD to support hostPID

olandr:
In order to leverage host-level things in a cluster, for instance when we want to access auditd or systemd files/processes, we need to enable the `hostPID` setting. This is by default not permitted in the otelcol CRDs.

This adds a second CRD version "v2beta1" with this field enabled. 

---------

on-behalf-of: @SAP Simon Olander (simon.olander@sap.com)

### DIFF
--- a/logs/charts/Chart.yaml
+++ b/logs/charts/Chart.yaml
@@ -4,7 +4,7 @@
 apiVersion: v2
 appVersion: v0.126.0
 name: opentelemetry-operator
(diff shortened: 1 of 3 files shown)


olandr:
I need this in this extension simply because we have not separated the operator from the chart yet into three parts.

Kuckkuck (approved):

PR #1056, opened 2025-07-25
fix(kube-monitoring) disable PrometheusSDRefreshFailure

trouaux:
alert a throwing absent metric on `prometheus_sd_refresh_failures_total`

this metric may only be present when actual sd config refresh are failing but greenhouse setup only uses selectors


viennaa (approved):
Thanks, now we're good :)

PR #1099, opened 2025-08-12
feat(thanos): resources spec on thanos store/compactor

trouaux:


viennaa (commented):
can we add ephemeral storage pls? to the store component, can default to 200Mi

PR #1177, opened 2025-09-01
fix(opensearch): improve logs-datastream policies

joluc:
## Pull Request Details

Improved OpenSearch logs-datastream policies with 7-day retention and basic authentication support for test environments.

- Updated ISM policy to 7-day retention instead of 30 days
- Changed from time-based to size-based rollover (10GB/20M docs) to avoid small indices
- Added basic auth configuration in `ci/test-values.yaml` for local development
- Fixed data stream template
- Simplify helpers/labels

## Breaking Changes

None. Changes are backwards compatible and only affect testing configuration in `ci/test-values.yaml`.

## Issues Fixed

1. Fixed OpenSearch Dashboards authentication for local testing environments
2. Fixed oversized indices by implementing proper size-based rollover thresholds

### DIFF
--- a/opensearch/chart/templates/dashboard-service.yaml
+++ b/opensearch/chart/templates/dashboard-service.yaml
@@ -4,6 +4,7 @@ kind: Service
 metadata:
   name: {{ .Values.cluster.cluster.name | default .Release.Name }}-dashboards-external
   labels:
+    {{- include "opensearch.labels" . | nindent 4 }}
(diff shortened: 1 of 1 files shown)


olandr (approved):
LGTM. I think the rationale for a shorter deletion deadline and a slimmer log rotation size is good. I guess one thing is to see how this progresses over time, if for instance we get plenty of load and the log rotation will keep chugging a long all the time (potentially taking resources to perform it).

PR #1200, opened 2025-09-25
feat(kube-monitoring): prometheus priority class

trouaux:


viennaa (commented):
I don't think we have Priority classes in greenhouse already.

This is the old chart:
https://github.com/sapcc/helm-charts/tree/master/system/priority-class

Supposedly this needs to go in before we can add this? Pls verify.

PR #1217, opened 2025-10-15
e2e flux

trouaux:


.github/workflows/e2e-flux-k8s.yaml
```diff
@@ -0,0 +1,167 @@
...
+          set -euo pipefail
+          chartfile=$(cat changed_charts.txt)
+          chart_dir="$(dirname "$chartfile")"
+          chart_yaml="$chart_dir/Chart.yaml"
+          [[ -f "$chart_yaml" ]] || { echo "No Chart.yaml at $chart_yaml"; exit 1; }
+          name="$(yq e '.name' "$chart_yaml")"
+          version="$(yq e '.version' "$chart_yaml")"
+          safe_name="$(echo "$name" | tr '[:upper:]' '[:lower:]' | tr -cs 'a-z0-9-' '-')"
+          echo "Applying HelmRelease for $name ($chart_dir)"
+          cat <<EOF | kubectl apply -f -
```
abhijith-darshan: Why not create a e2e folder and maintain all these test case inline yamls in there? It will be useful to replicate any flaky tests locally. And your workflow gets small.

.github/workflows/license.yaml
```diff
@@ -6,7 +6,7 @@ on:
 jobs:
   generate-token:
     name: Generate GitHub App Token
-    runs-on: default
+    runs-on: none
```
abhijith-darshan: This would be invalid. revert to default

PR #1220, opened 2025-10-15
feature: adding keystone_scope parsing for ceph/prysm-sidecar container

Kuckkuck:
Adding keystone_scope parsing for ceph/prysm-sidecar container


kayrus (commented):
please add these fields as well

kayrus (approved):
LGTM, thanks!

PR #972, opened 2025-06-17
feat(extensions): migrate from PluginDefinition to ClusterPluginDefinition

abhijith-darshan:
## Pull Request Details

This PR migrates `PluginDefinition` CR to `ClusterPluginDefinition` 

This is due to the enhancement of `PluginDefinition` which is being transitioned from cluster scoped CR to namespaced scoped CR.

The issue is documented here https://github.com/cloudoperators/greenhouse/issues/1004 and here https://github.com/cloudoperators/greenhouse/issues/1214

## Breaking Changes

- While this is breaking change, greenhouse controllers ensure smooth transition without any user action necessary

PR #1233, opened 2025-10-24
chore(all): Migrate back to PD

uwe-mayer:
This is the second part of https://github.com/cloudoperators/greenhouse-extensions/pull/972

When migrating from cluster scoped PluginDefinitions to 
- [ClusterPluginDefinitions](https://github.com/cloudoperators/greenhouse/issues/1214) (cluster scope)
- [PluginDefinitions](https://github.com/cloudoperators/greenhouse/issues/1004) (ns scope)

We first moved every PD to CPD for a non-breaking transition and now move everything back to PD with the release and rollout of the new feature.

Please do not merge!


dimtass (approved):
Alles PluginDefinition

PR #1259, opened 2025-11-12
feat(kafka): initial helm chart

joluc:
## Pull Request Details

Adds a new Kafka plugin for Greenhouse using the Strimzi Kafka Operator with KRaft mode (without ZooKeeper).

### DIFF
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -15,8 +15,6 @@ The **Strimzi Kafka Operator** simplifies the management of Kafka clusters on Ku
 
 - [Strimzi Kafka Operator](https://github.com/strimzi/strimzi-kafka-operator)
 - Apache Kafka Cluster Management (KRaft mode)
-- Kafka Connect (optional)
-- Kafka Bridge (optional)
 - Kafka Exporter for Metrics (optional)
 - Cruise Control for Cluster Optimization (optional)
 - Entity Operator for Topic and User Management
--- a/kafka/README.md.gotmpl
+++ b/kafka/README.md.gotmpl
@@ -15,8 +15,6 @@ The **Strimzi Kafka Operator** simplifies the management of Kafka clusters on Ku
 
 - [Strimzi Kafka Operator](https://github.com/strimzi/strimzi-kafka-operator)
 - Apache Kafka Cluster Management (KRaft mode)
-- Kafka Connect (optional)
-- Kafka Bridge (optional)
 - Kafka Exporter for Metrics (optional)
 - Cruise Control for Cluster Optimization (optional)
 - Entity Operator for Topic and User Management


kafka/chart/templates/kafka-topics.yaml
```diff
@@ -0,0 +1,52 @@
...
+    cleanup.policy: {{ $topic.cleanupPolicy | default "delete" }}
+
+    # Max message size
+    {{- if $topic.maxMessageBytes }}
+    max.message.bytes: {{ $topic.maxMessageBytes }}
+    {{- end }}
+
+    {{- with $topic.additionalConfig }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
```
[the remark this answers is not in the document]
joluc: Example from [here](https://github.com/strimzi/strimzi-kafka-operator/blob/main/examples/topic/kafka-topic.yaml), structure and description defined in [stimzi docs](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaTopic-reference) added config from [Apache Kafka docs](https://kafka.apache.org/documentation/#topicconfigs) and successfully tested in my kind cluster :)

kafka/chart/templates/service-monitor.yaml
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and Greenhouse contributors
+# SPDX-License-Identifier: Apache-2.0
+
+{{- if and .Values.kafka.enabled .Values.monitoring.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ .Values.kafka.name }}-kafka
```
olandr: In the values.yaml below the .`Values.kafka.name = kafka-cluster`. So 'kafka' is both prepended and appended by default.

kafka/chart/values.yaml
```diff
@@ -0,0 +1,210 @@
...
+      cpu: "1000m"
+    limits:
+      memory: "4Gi"
+      cpu: "2000m"
+
+  # -- JVM options for Kafka brokers
+  # @default -- xms: 1024m, xmx: 2048m
+  jvmOptions:
+    xms: "1024m"
+    xmx: "2048m"
```
olandr: The @'defaults are good hints, but I do not think it gives a good overview of why these are defaults.

As this is a java app, I am _guessing_ that these are not "defaults" but rather minimum requirements. Is that the case?

kafka/chart/templates/service-monitor.yaml
timojohlo: valid point. I have changed it to `kafka-metrics`

kafka/chart/values.yaml
joluc: I've extended the comment a bit. We need to tweak around with those values. Good read is https://www.automq.com/blog/kafka-design-page-cache-performance

kafka/chart/templates/metrics-configmap.yaml
```diff
@@ -0,0 +1,64 @@
...
+      name: kafka_$1_$2_$3_percent
+      type: GAUGE
+    - pattern: kafka.(\w+)<type=(.+), name=(.+)Percent\w*><>Value
+      name: kafka_$1_$2_$3_percent
+      type: GAUGE
+    - pattern: kafka.(\w+)<type=(.+), name=(.+)Percent\w*, (.+)=(.+)><>Value
+      name: kafka_$1_$2_$3_percent
+      type: GAUGE
+      labels:
+        "$4": "$5"
```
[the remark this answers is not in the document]
joluc: Do avoid conflict I added a suffix on the rate one and this works like a charme :)

olandr (approved):
Looks good to me!

IvoGoman (approved):

PR #1363, opened 2026-01-13
feat(.github): speedup docker by removing arm64, adds cache

olandr:
x6 speed improvement of the docker logs build by removing the arm64 build from the step.

Also removed the arm64 (through templating just the amd64 instead) of the integration test

---------


timojohlo (approved):

ibakshay (approved):
LGTM!

PR #1456, opened 2026-03-02
feat(kube-monitoring)!: make Perses datasource default configurable

ibakshay:
## Description
- Add configurable isDefault option for Perses datasource in kube-monitoring chart
- Follow the same pattern used in the thanos chart

## Breaking Change
This PR introduces a breaking change to the kubeMonitoring.prometheus.persesDatasource configuration.

### Before:
```yaml
kubeMonitoring:
  prometheus:
    persesDatasource: true
```

### After:

```yaml
kubeMonitoring:
  prometheus:
    persesDatasource:
      create: true
      isDefault: false
```

`persesDatasource.create: true` has the same behaviour as before and it is not breaking.

### DIFF
--- a/kube-monitoring/charts/values.yaml
+++ b/kube-monitoring/charts/values.yaml
@@ -345,7 +345,7 @@ kubeMonitoring:
     # @ignored
     persesDatasource:
       create: true
-      isDefault: false 
+      isDefault: false
 
     prometheusSpec:
 


viennaa (commented):
What happens with this dashboard? Is this conflicting?


https://github.com/cloudoperators/greenhouse-extensions/blob/main/kube-monitoring/charts/perses-dashboards/prometheus.json

viennaa (approved):

PR #1458, opened 2026-03-04
feat(thanos): expose queryEndpoints and alertQueryUrl; fix ruleSelector nindent bug

kevin-fischer:
## Problem

ThanosRuler in chart v0.8.0 has three bugs:

1. **ruleSelector nindent 4 bug** - labels render as siblings of matchLabels, K8s ignores them, resulting in empty selector
2. **queryEndpoints hardcoded** - cannot configure regional query
3. **alertQueryUrl missing** - no alert links in Alertmanager

## Solution

- Fix nindent 4 to 6 for proper label nesting under matchLabels
- Add queryEndpoints if/else logic with local service fallback
- Add alertQueryUrl conditional rendering

Validated against prometheus-operator CRD spec. Backward compatible.


viennaa (commented):
Thanks for bringing up the PR.

Some small nits before we can go live.

PR #1460, opened 2026-03-04
Feat/thanos add store flags

sjentzschsap:
## Pull Request Details

Allow defining additional stores for thanos query without requiring to also re-list the default ones

## Breaking Changes

none

## Issues Fixed

SAP-internal tickets; no need to post them here I guess

## Other Relevant Information

As aligned


trouaux (changes requested):
seems unnecessary, can you describe a scenario where you would benefit from this?

viennaa:
I've aligned with @sjentzschsap on how to continue here. We're going to leverage a stand alone query for incorporating the other queries, so this PR is currently no blocker.

However, in a upstream world this PR makes sense. People could decide to add more targets to their first deployed query in addition to the sidecar and store which is configured by default. If we don't allow this, everyone is always forced to spin up a new query to achieve a multiple layer architecture.

PR #1543, opened 2026-04-10
fix(opensearch): disable drain on rolling restart

joluc:
## Pull Request Details

With replicas enabled, draining is unnecessary. Replicas are promoted to primaries when a node restarts, and replica shards are reassigned when it rejoins. Draining causes the operator to get stuck waiting for green status during rolling restarts. Hence, disable it by default for both logs and SIEM clusters.


Kuckkuck (approved):

joluc:
Covered in https://github.com/cloudoperators/greenhouse-extensions/pull/1568

PR #1546, opened 2026-04-10
chore(ci): bump plugindefinition version.spec on regex manager

trouaux:
enable plugindefinition version bump 

eg. https://github.com/cloudoperators/greenhouse-extensions/pull/1539


ibakshay (commented):
Are you aware that `Automated plugin README generation` with updated values will be dropped as well? shall we have a thin workflow for that?

ibakshay (approved):
Thank you! LGTM!

PR #1572, opened 2026-04-28
feat(gatekeeper): add PluginDefinition for OPA Gatekeeper

### LINKED ISSUE #1420: [gatekeeper] First release
### Priority

(Medium) I'm annoyed but I'll live

### Is your feature request related to a problem?

With managing a fleet of clusters we want to provide a Greenhouse PD that allows configuring admission policies on remote clusters. The target is to provide the [already existing helm chart for gatekeeper](https://github.com/sapcc/helm-charts/tree/master/system/gatekeeper) used in SAP Cloud Infrastructure as a Greenhouse PD.

With this issue we want to have a first release of the PD that ships the operator and a good understanding on how to ship dependent configuration. Currently a helm post-install hook is used. 
With this first iteration we can keep this hook, but want to explore if publishing a depending gatekeeper config PD makes sense

Acceptance criteria:

- [ ] The existing gatekeeper helm chart is configurable and deployable as a Greenhouse PluginDefinition.
- [ ] If decided upon a ticket exists for a depending PD shipping gatekeeper config.

### Additional context.

_No response_

### Reference Issues

_No response_

mikolajkucinski:
## Pull Request Details

Adds a new `gatekeeper` PluginDefinition to the catalog, wrapping the upstream OPA Gatekeeper Helm chart.

## Breaking Changes

None. This is a new PluginDefinition and does not modify any existing ones.

## Issues Fixed

1. Closes #1420

## Other Relevant Information

See the README for details. A follow-up ticket will track the `gatekeeper-config` PluginDefinition that ships `ConstraintTemplates` and `Constraints` originating from the legacy `sapcc/helm-charts/system/gatekeeper` chart.


majewsky:
Comparing against the existing implementation in https://github.com/sapcc/helm-charts/tree/master/system/gatekeeper:

1. The current chart engages in vendoring shenanigans to ensure that `linkerd.io/inject` annotations are added on all pods. Has this been considered? I'm not saying that the existing mechanism needs to be copied over verbatim, but since there is no mention of linkerd in the PR as of right now, I want to flag this as a concern. The existing chart explicitly disables linkerd injection because the only relevant network traffic goes through k8s apiserver or ingress and thus is already encrypted, so it is possible that the answer is just "no linkerd annotations needed here".

2. Since DOOP (Decentralized Observer Of Policies, see docu for [doop-analyzer](https://github.com/sapcc/gatekeeper-addons/tree/main/doop-analyzer) and [doop-api](https://github.com/sapcc/gatekeeper-addons/tree/main/doop-analyzer)) is an integral part of a Gatekeeper multi-cluster setup, I feel like it ought to be part of this plugin. I'm not sure how this connects to the DOOP UI, which is a different Greenhouse plugin, so I'm only flagging this for consideration.

3. I suggest reviewing https://github.com/sapcc/helm-charts/blob/master/system/gatekeeper/values.yaml for additional values that need to be set. For example, `disabledBuiltins = []` will be necessary because policies using helm-manifest-parser or doop-image-checker need `http.send()`.

mikolajkucinski:
> Comparing against the existing implementation in https://github.com/sapcc/helm-charts/tree/master/system/gatekeeper:
> 
> 1. The current chart engages in vendoring shenanigans to ensure that `linkerd.io/inject` annotations are added on all pods. Has this been considered? I'm not saying that the existing mechanism needs to be copied over verbatim, but since there is no mention of linkerd in the PR as of right now, I want to flag this as a concern. The existing chart explicitly disables linkerd injection because the only relevant network traffic goes through k8s apiserver or ingress and thus is already encrypted, so it is possible that the answer is just "no linkerd annotations needed here".
> 2. Since DOOP (Decentralized Observer Of Policies, see docu for [doop-analyzer](https://github.com/sapcc/gatekeeper-addons/tree/main/doop-analyzer) and [doop-api](https://github.com/sapcc/gatekeeper-addons/tree/main/doop-analyzer)) is an integral part of a Gatekeeper multi-cluster setup, I feel like it ought to be part of this plugin. I'm not sure how this connects to the DOOP UI, which is a different Greenhouse plugin, so I'm only flagging this for consideration.
> 3. I suggest reviewing https://github.com/sapcc/helm-charts/blob/master/system/gatekeeper/values.yaml for additional values that need to be set. For example, `disabledBuiltins = []` will be necessary because policies using helm-manifest-parser or doop-image-checker need `http.send()`.

Thanks Stefan, addressing each:

1. No linkerd annotations are set by default. The upstream chart 3.22.2 exposes `podAnnotations` and `auditPodAnnotations` so deployments running Linkerd can disable injection via PluginPreset:

```yaml
optionValues:
  - name: podAnnotations
    value:
      linkerd.io/inject: disabled
  - name: auditPodAnnotations
    value:
      linkerd.io/inject: disabled
```

2. Per discussion with @IvoGoman and @uwe-mayer, I propose to:
- Keep the gatekeeper operator decoupled from DOOP
- Ship DOOP across separate PDs (mirroring the legacy `system/gatekeeper` + `system/doop-central` split, but with per-cluster DOOP components extracted into their own PD instead of bundled into the operator):
  - `gatekeeper-doop` - per-cluster components (doop-analyzer, doop-image-checker, helm-manifest-parser)
  - `doop-central` - centralized doop (api + landing-page + ingress)
  - `doop` - existing UI plugin

3. Good catch, you're right about `http.send()`. Upstream 3.22.2 has `disabledBuiltins: ["{http.send}"]` by default (http.send disabled), so policies using `helm-manifest-parser` or `doop-image-checker` would fail without an override. I'll add `disabledBuiltins` as a configurable option and also sweep for others worth surfacing.

gatekeeper/plugindefinition.yaml
```diff
@@ -0,0 +1,74 @@
...
+  description: |
+    Policy controller for Kubernetes admission based on the Open Policy Agent
+    (OPA) constraint framework. Deploys upstream OPA Gatekeeper. Constraint
+    templates and constraints are not bundled; install them separately or
+    use the gatekeeper-config PluginDefinition (planned).
+  docMarkDownUrl: https://raw.githubusercontent.com/cloudoperators/greenhouse-extensions/main/gatekeeper/README.md
+  icon: https://raw.githubusercontent.com/open-policy-agent/gatekeeper/master/logo/gatekeeper-horizontal-color.png
+  helmChart:
+    name: gatekeeper
+    repository: https://open-policy-agent.github.io/gatekeeper/charts
```
abhijith-darshan: We need OCI here otherwise chart replication cannot be triggered

gatekeeper/charts/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+icon: https://raw.githubusercontent.com/open-policy-agent/gatekeeper/master/logo/gatekeeper-horizontal-color.png
+keywords:
+  - gatekeeper
+  - opa
+  - policy
+  - admission-control
+version: 1.0.0
+appVersion: 3.22.2
+maintainers:
+  - name: ivogoman
```
abhijith-darshan: You want to put yourself here instead of ivogoman?

mikolajkucinski: As discussed, updating that list :)

uwe-mayer (approved):

PR #1598, opened 2026-05-07
fix(kube-monitoring): incorporate PrometheusRule with thanos-ruler label for absent-metrics-operator

sjentzschsap:
## Pull Request Details

The absent-metrics-operator logs errors and has issues processing PrometheusRules that include `thanos-ruler` labels, and neither `plugin` nor `prometheus`. This is the case, when PrometheusRules are processed by Thanos Rule.

According [to the docs](https://github.com/cloudoperators/absent-metrics-operator/blob/master/docs/absence-alert-rule-definition.md#aggregation) this is already the default, so let's not deviate here, and respect the `thanos-ruler` labels here too if present.

### DIFF
--- a/kube-monitoring/plugindefinition.yaml
+++ b/kube-monitoring/plugindefinition.yaml
@@ -15,7 +15,7 @@ spec:
     # renovate depName=ghcr.io/cloudoperators/greenhouse-extensions/charts/kube-monitoring
     name: kube-monitoring
     repository: oci://ghcr.io/cloudoperators/greenhouse-extensions/charts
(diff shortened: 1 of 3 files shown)


viennaa (approved):
Solid PR, thanks 👍

kube-monitoring/plugindefinition.yaml
viennaa: we need this bumped as well as the kube-monitoring chart.

sjentzschsap: done

PR #1626, opened 2026-05-18
feat(thanos) add thanos-community helm charts

trouaux:


thanos/charts/Chart.yaml
```diff
@@ -11,8 +11,10 @@ maintainers:
...
-keywords:
-  - thanos
-  - storage
-  - metrics
+version: 0.9.1
+dependencies:
+  - name: thanos
+    repository: https://thanos-community.github.io/helm-charts/
+    version: 0.11.0
+    alias: thanosV1
```
viennaa:
```suggestion
    alias: thanosCommunity
```

Would try to make it a bit clearer what V1 is. We need to transition off this a bit later.

viennaa (approved):

PR #1627, opened 2026-05-18
feat(thanos): add per-alert disable and label override support

jeoffzheng:
Provide the capability/option for downstream(KVM cluster) to disable the alerts/overwrite lables 

### summary of changes 
  - Add `thanos.serviceMonitor.disabledAlerts` — list of alert names to suppress entirely from rendered PrometheusRule objects
  - Add `thanos.serviceMonitor.alertOverrides` — map of alert name → label overrides (e.g. override severity, add custom labels)
- Refactor alerts.yaml to parse alert YAML via fromYaml for rule-level filtering
- Expose both options in plugindefinition.yaml as list/map options
- Bump chart and plugin version 0.9.0 → 0.10.0

### DIFF
--- a/thanos/charts/Chart.yaml
+++ b/thanos/charts/Chart.yaml
@@ -11,10 +11,14 @@ maintainers:
 name: thanos
 sources:
   - https://github.com/cloudoperators/greenhouse-extensions
-version: 0.9.1
+version: 0.10.0
 dependencies:
   - name: thanos
     repository: https://thanos-community.github.io/helm-charts/
     version: 0.11.0
     alias: thanosCommunity
     condition: thanosCommunity.enabled
+keywords:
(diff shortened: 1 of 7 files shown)


thanos/charts/Chart.yaml
```diff
@@ -11,10 +11,14 @@ maintainers:
...
 dependencies:
   - name: thanos
     repository: https://thanos-community.github.io/helm-charts/
     version: 0.11.0
     alias: thanosCommunity
     condition: thanosCommunity.enabled
+keywords:
+  - thanos
+  - storage
+  - metrics
```
trouaux, line 24:
```suggestion
```

trouaux (approved):

PR #1654, opened 2026-05-21
fix(plugindefinition): add missing required fields to alerts and disco options

onuryilmaz:
## Summary

- Adds missing `required: false` to `alerts.dashboard.create` in `alerts/plugindefinition.yaml`
- Adds missing `required: false` to `image.repository` in `disco/plugindefinition.yaml`

## Context

The `PluginDefinition` CRD schema enforces that every entry in `spec.options` must have the `required` field explicitly set (the Go struct tag has no `omitempty`). The kustomize-controller performs a server-side apply dry-run on every reconciliation, which runs full API server validation. An option missing `required:` causes the dry-run to fail with a `Required value` validation error.

This was surfaced by the greenhouse catalog E2E test after upgrading to a newer kustomize-controller version that enforces strict server-side dry-run validation. See: https://fluxcd.io/flux/components/kustomize/kustomizations/#server-side-apply

## Test plan

- [ ] Confirm the greenhouse catalog E2E passes after merging


trouaux:
could you bump respective charts plugindefintions versions

onuryilmaz:
> could you bump respective charts plugindefintions versions

done with https://github.com/cloudoperators/greenhouse-extensions/pull/1654/commits/4ff9e778ad2379502c0b2a09badaa16536cdecd9

trouaux (approved):

IvoGoman (approved):

PR #1682, opened 2026-06-02
feat(gatekeeper-config): foundation chart with two universal policies

mikolajkucinski:
## Pull Request Details

Add the `gatekeeper-config` PluginDefinition skeleton plus two universal admission policies. This is the foundation of a larger migration of OPA Gatekeeper policies into a Greenhouse PluginDefinition. 

**Differences vs upstream:**

| Policy / Lib | Upstream | This PR |
|---|---|---|
| `high-cpu-requests` | Only `containers` and silently drops malformed CPU values. | Includes `initContainers` and explicit violation on unparseable CPU. |
| `unmanaged-pods` | Hardcoded `^recycler-for-` exemption in `default` namespace. | `allowedNamePatterns` parameter (regex to prevent prefix bypass) |
| `_helpers.tpl::add_support_labels` | Writes `ccloud/support-group` + `ccloud/service` into messages. | Writes single `greenhouse.sap/owned-by`. |

## Breaking Changes

None. This adds a new PluginDefinition.

## Issues Fixed

- Partial progress on [#1420](https://github.com/cloudoperators/greenhouse-extensions/issues/1420)

## Other Relevant Information

None.

### DIFF
--- a/.github/workflows/gatekeeper-config-tests.yaml
+++ b/.github/workflows/gatekeeper-config-tests.yaml
@@ -5,6 +5,11 @@ on:
     paths:
       - gatekeeper-config/**
       - .github/workflows/gatekeeper-config-tests.yaml
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
 
 jobs:
   gator-verify:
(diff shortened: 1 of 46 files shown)


majewsky (dismissed):
This is missing the test suite from the existing Gatekeeper chart. This is an absolutely vital piece of the puzzle. From experience, you will have an atrociously miserable time maintaining these policies without a good test suite that you can run without having to go through a (multi-stage in this case!) Helm deployment.

Explicitly putting this as a red review because figuring out how to make the test suite work in this repo is (from my PoV) the most important part of the DoD for this step.

abhijith-darshan (approved):

uwe-mayer (approved):

PR #1715, opened 2026-06-10
feat(opensearch): build opensearch-fortlogs

ztomaszewska:
## Pull Request Details
We need to build our own custom OpenSearch image for Fortlogs solution. It's based on official OpenSearch image. Two official plugins are removed and replaced with custom ones (that include some critical fixes). 
It's a workaround for limitation of opensearch k8s operator (https://github.com/opensearch-project/opensearch-k8s-operator/issues/1423).


ztomaszewska:
Closing. - replace with this: https://github.com/cloudoperators/greenhouse-extensions/pull/1757

PR #1720, opened 2026-06-10
feat(logs): add standalone external collector for devices

timojohlo:
## Details
Introduces a standalone OpenTelemetry external collector (StatefulSet) that is independent from the existing logs DaemonSet collector. This new collector handles log ingestion from external devices — syslog (TCP/UDP/TLS) and traces (OTLP gRPC/HTTP) — and will later support routing to different kafka topics (audit, operational) based on defined attributes.

### Changes
- New external collector StatefulSet (`external-collector.yaml`) with dedicated PodMonitor for Prometheus scraping
- Extracted syslog config into reusable `_syslog-config.tpl` helper template
- Restructured values: moved Kafka settings to a shared openTelemetry.kafka block with per-collector topic sub-keys; moved syslog/traces/external configs under openTelemetry.externalCollector
- Service type change: external service switched from NodePort + externalIPs to LoadBalancer with configurable loadBalancerIP
- Conditional secret: otel-basic-auth secret only rendered when Kafka is disabled (OpenSearch failover mode)
- Removed external ingestion from DaemonSet: the logs collector no longer handles syslog, traces, or external webhook receivers


logs/charts/templates/external-service.yaml
```diff
@@ -1,59 +1,58 @@
...
     app.kubernetes.io/component: opentelemetry-collector
-    app.kubernetes.io/instance: {{ .Release.Namespace }}.logs
+    app.kubernetes.io/instance: {{ .Release.Namespace }}.external
     app.kubernetes.io/managed-by: opentelemetry-operator
-    app.kubernetes.io/part-of: opentelemetry
   type: NodePort
-  externalTrafficPolicy: Local
-{{- if .Values.openTelemetry.logsCollector.externalConfig.external_ip }}
-  externalIPs: ["{{ .Values.openTelemetry.logsCollector.externalConfig.external_ip }}"]
+  externalTrafficPolicy: Cluster
```
joluc: Why `Cluster`?

logs/charts/ci/test-values-all-features.yaml
```diff
@@ -52,15 +52,20 @@ openTelemetry:
       enabled: true
     openstackConfig:
       enabled: true
+  externalCollector:
+    enabled: false
```
joluc: `false` but all with all sub-toggles `true`? Shouldn't we test also the externalCollector in the all features test?

logs/charts/values.yaml
```diff
@@ -109,6 +105,32 @@ openTelemetry:
...
+  # -- Standalone external OTel Collector as StatefulSet.
+  externalCollector:
+    # -- Enables the standalone external OTel Collector StatefulSet and its PodMonitor.
+    enabled: false
+    # -- External IP exposed on the NodePort service
+    external_ip:
+    # -- Additional annotations on the external Service
+    serviceAnnotations: {}
+    # -- Kafka exporter configuration for the external collector
+    kafka:
```
joluc: Actually this duplicates `logsCollector.kafka`, in practice both collectors run in the same cluster and target the same Kafka. We would set the settings twice (no single source of truth). Maybe we use a shared openTelemetry.kafka block?

logs/charts/templates/external-service.yaml
timojohlo: The intention was that if we switch to `StatefulSet` we also need to make sure we can see the endpoints cluster-wide and not only per-node.

logs/charts/ci/test-values-all-features.yaml
timojohlo: addressed

logs/charts/values.yaml
timojohlo: addressed

Kuckkuck (approved):

logs/charts/ci/test-values-all-features.yaml
```diff
@@ -36,10 +36,10 @@ openTelemetry:
     failover_password_b: test
   cluster: test
   region: test
+  kafka:
+    enabled: true
   logsCollector:
```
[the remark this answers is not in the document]
timojohlo: addressed

logs/charts/templates/syslog-tls-certificate.yaml
```diff
@@ -2,7 +2,7 @@
 SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Greenhouse contributors
 SPDX-License-Identifier: Apache-2.0
 */}}
-{{- if .Values.openTelemetry.logsCollector.syslogTLSConfig.enabled }}
+{{- if .Values.openTelemetry.externalCollector.syslogTLSConfig.enabled }}
```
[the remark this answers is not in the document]
timojohlo: addressed

logs/charts/templates/external-service.yaml
```diff
@@ -1,59 +1,58 @@
-{{- if or .Values.openTelemetry.logsCollector.externalConfig.enabled .Values.openTelemetry.logsCollector.syslogConfig.enabled .Values.openTelemetry.logsCollector.syslogTLSConfig.enabled .Values.openTelemetry.logsCollector.tracesConfig.enabled }}
+{{- if .Values.openTelemetry.externalCollector.enabled }}
```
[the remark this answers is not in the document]
timojohlo: addressed

logs/charts/templates/logs-collector.yaml
```diff
@@ -36,76 +38,29 @@ spec:
...
-          fieldPath: spec.nodeName
-    - name: MY_POD_IP
-      valueFrom:
-        fieldRef:
-          apiVersion: v1
-          fieldPath: status.podIP
     - name: cluster
       value: "{{ .Values.openTelemetry.cluster }}"
     - name: region
       value: "{{ .Values.openTelemetry.region }}"
```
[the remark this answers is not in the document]
timojohlo: addressed

logs/charts/templates/secret.yaml
```diff
@@ -17,4 +18,5 @@ data:
    failover_password_a: {{ .Values.openTelemetry.openSearchLogs.failover_password_a | b64enc | quote }}
    failover_username_b: {{ .Values.openTelemetry.openSearchLogs.failover_username_b | b64enc | quote }}
    failover_password_b: {{ .Values.openTelemetry.openSearchLogs.failover_password_b | b64enc | quote }}
```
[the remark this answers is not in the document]
timojohlo: required would mean it should be set even if we ship logs to Kafka. That does not make sense.

joluc (approved):

PR #1757, opened 2026-06-23
feat(opensearch): build opensearch-fortlogs

ztomaszewska:
## Pull Request Details

We need to build our own custom OpenSearch image for Fortlogs solution. It's based on official OpenSearch image. Two official plugins are removed and replaced with custom ones (that include some critical fixes).
It's a workaround for limitation of opensearch k8s operator (https://github.com/opensearch-project/opensearch-k8s-operator/issues/1423).


joluc (changes requested):
I would keep the Dockerfile [here](https://github.com/SAP-cloud-infrastructure/opensearch-fortlogs) and reference it in the Context like [here](https://github.com/cloudoperators/greenhouse-extensions/blob/a936c30b685fec395a435d5de167a507e135a64f/.github/workflows/docker-build.yaml#L34)

ztomaszewska:
> I would keep the Dockerfile [here](https://github.com/SAP-cloud-infrastructure/opensearch-fortlogs) and reference it in the Context like [here](https://github.com/cloudoperators/greenhouse-extensions/blob/a936c30b685fec395a435d5de167a507e135a64f/.github/workflows/docker-build.yaml#L34)

Changes done - Dockerfile back again [here](https://github.com/SAP-cloud-infrastructure/opensearch-fortlogs/blob/main/Dockerfile)

In order to make it work `opensearch-fortlogs` repo should be public.

joluc (approved):

PR #1840, opened 2026-07-22
feat(logs): add SLO burn rate, export failure, queue, and volume alerts

wiltongarcia:
## Summary

- Add \`OTelLogsSLOBurnRateCritical\` + \`OTelLogsSLOBurnRateWarning\` — multi-window burn rate alerts (14x/6x) for the 99.5% export SLO
- Add \`OTelLogsExportFailureRatioHigh\` — fast per-exporter export failure detection (10m window, 5m \`for:\`) replacing the slow 2h \`OTelLogsExportingFailed\`
- Add \`OTelExporterQueueFull\` — pipeline stalled alert (queue at 100% capacity with \`block_on_overflow=true\`)
- The following alerts are available but **disabled by default** (visible on dashboard, not paging): \`OTelExporterQueueFilling\`, \`OTelLogsVolumeSpike\`, \`OTelLogsVolumeDropFast\`
- Add playbook stubs for all new alerts
- Bump chart to \`0.4.47\`, plugindefinition to \`0.15.47\`

All **new** alerts use \`severity: info\` pending production validation. Pre-existing alerts (\`OTelLogsMissing\`, \`ReconcileErrors\`, \`WorkqueueDepth\`, \`ReceiverRefusedMetric\`) retain their original \`severity: warning\`.


logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +126,220 @@ spec:
...
+      - alert: OTelLogsSLOBurnRateWarning
+        expr: |
+          (
+            sum by (k8s_cluster_name) (rate(otelcol_exporter_send_failed_log_records_total{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}[6h]))
+            /
+            (
+              sum by (k8s_cluster_name) (rate(otelcol_exporter_sent_log_records_total{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}[6h]))
+              + sum by (k8s_cluster_name) (rate(otelcol_exporter_send_failed_log_records_total{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}[6h]))
+            )
+          ) > (6 * 0.005)
```
olandr, line 191: I am not understanding where this comes from `6 * 0.005`? What unit and purpose does 0.005 have here?

Like, the calculation would never be 6.00 so I get that it is less than 1. But I cannot figure out why it is scaled by 0.005 specifically 🤔

logs/charts/templates/alerts.yaml
```diff
@@ -106,7 +104,6 @@ spec:
         expr: sum(rate(otelcol_exporter_sent_log_records_total{job=~".*/opentelemetry-collector-logs"}[1h]offset 2h)) by (k8s_cluster_name)/sum(rate(otelcol_exporter_sent_log_records_total{job=~".*/opentelemetry-collector-logs"}[1h])) by (k8s_cluster_name) > 4
         for: 2h
         labels:
-          context: logshipping
```
joluc, line 109: We don't use the context for routing but I checked again and it seems we are using it still for inhibition rules (region/cluster/context). Maybe let's keep it then

logs/charts/templates/alerts.yaml
wiltongarcia: I am setting an initial SLO to 99.5%, so the acceptable failure rate is 0.5% = 0.005.

The SLO should be consumed in 28 days (1.0), the 6 is used to see if the SLO is been consumed 6x fast than we expect, 28 / 6 ≈ 4.5 days. So 6 * 0.005 = 0.03 is simply the threshold: if the current error ratio exceeds 3%, you are by definition burning at 6× rate. 

So we will alert only if we fail faster that we expect, and the 6x will be a warning and the 14x will be critical, but for now as we are testing, all severities are set for warning.

olandr: I see, 6x5‰ is the rate of consumption to trigger on. This would then mean that if 3% of failures happen in any 6h window, we get a warning. Correct? 7% for the 14x critical.

A very low threshold but I guess that is what you need with 5‰ error budgets :)

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+            /
+            (
+              sum by (k8s_cluster_name) (rate(otelcol_exporter_sent_log_records_total{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}[5m]))
+              + sum by (k8s_cluster_name) (rate(otelcol_exporter_send_failed_log_records_total{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}[5m]))
+            )
+          ) > (14 * 0.005)
+        for: 2m
+        labels:
+          context: logshipping
+          severity: warning
```
timojohlo:
```suggestion
          severity: info
```
lets set this to `info` first and gain some confidence.

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+            /
+            (
+              sum by (k8s_cluster_name) (rate(otelcol_exporter_sent_log_records_total{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}[30m]))
+              + sum by (k8s_cluster_name) (rate(otelcol_exporter_send_failed_log_records_total{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}[30m]))
+            )
+          ) > (6 * 0.005)
+        for: 15m
+        labels:
+          context: logshipping
+          severity: warning
```
timojohlo:
```suggestion
          severity: info
```
lets set this to `info` first and gain some confidence.

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+              rate(otelcol_exporter_send_failed_log_records_total{
+                job=~".*/opentelemetry-collector-logs",
+                exporter!~"debug"
+              }[10m])
+            )
+          ) > 0.1
+        for: 5m
+        labels:
+          context: logshipping
+          severity: warning
```
timojohlo:
```suggestion
          severity: info
```
lets set this to `info` first and gain some confidence.

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+        expr: |
+          (
+            otelcol_exporter_queue_size{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}
+            /
+            otelcol_exporter_queue_capacity{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}
+          ) > 0.8
+        for: 5m
+        labels:
+          context: logshipping
+          severity: warning
```
timojohlo:
```suggestion
          severity: info
```
lets set this to `info` first and gain some confidence.

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+        expr: |
+          (
+            otelcol_exporter_queue_size{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}
+            /
+            otelcol_exporter_queue_capacity{job=~".*/opentelemetry-collector-logs", exporter!~"debug"}
+          ) >= 1.0
+        for: 2m
+        labels:
+          context: logshipping
+          severity: warning
```
timojohlo:
```suggestion
          severity: info
```
lets set this to `info` first and gain some confidence.

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+          /
+          sum by (k8s_cluster_name) (
+            rate(otelcol_receiver_accepted_log_records_total{
+              job=~".*/opentelemetry-collector-logs"
+            }[1h] offset 1h)
+          ) > 5
+        for: 5m
+        labels:
+          context: logshipping
+          severity: warning
```
timojohlo:
```suggestion
          severity: info
```
lets set this to `info` first and gain some confidence.

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+          sum by (k8s_cluster_name) (
+            rate(otelcol_exporter_sent_log_records_total{
+              job=~".*/opentelemetry-collector-logs",
+              exporter!~"debug"
+            }[1h] offset 1h)
+          ) < 0.2
+        for: 10m
+        labels:
+          context: logshipping
+          severity: warning
```
timojohlo:
```suggestion
          severity: info
```
lets set this to `info` first and gain some confidence.

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+          description: 'OTel on {{`{{ $labels.k8s_cluster_name }}`}} is exporting less than 20% of its 1-hour baseline volume. Detected within 10 minutes vs the previous 4-hour detection time.'
+  {{- end }}
+
+  {{- if (has "JournaldReceiverRefusedLogs" .Values.openTelemetry.prometheus.rules.enabled) }}
+      - alert: JournaldReceiverRefusedLogs
+        expr: sum(rate(otelcol_receiver_refused_log_records_total{receiver=~"journald"}[1m])) > 0
+        for: 5m
+        labels:
+          context: logshipping
+          severity: warning
```
timojohlo:
```suggestion
          severity: info
```
lets set this to `info` first and gain some confidence.

logs/charts/templates/alerts.yaml
```diff
@@ -92,7 +92,7 @@ spec:
         for: 6h
         labels:
           context: logshipping
-          severity: info
+          severity: warning
```
timojohlo: is there a reason to raise the severity here?

logs/charts/templates/alerts.yaml
```diff
@@ -130,4 +130,229 @@ spec:
...
+          playbook: https://github.com/cloudoperators/greenhouse-extensions/tree/main/logs/playbooks/OTelLogsVolumeAnomaly.md
+          {{- include "plugin.additionalRuleLabels" . | nindent 10 }}
+        annotations:
+          summary: 'Log volume dropped >80% vs 1h baseline on {{`{{ $labels.k8s_cluster_name }}`}}'
+          description: 'OTel on {{`{{ $labels.k8s_cluster_name }}`}} is exporting less than 20% of its 1-hour baseline volume. Detected within 10 minutes vs the previous 4-hour detection time.'
+  {{- end }}
+
+  {{- if (has "JournaldReceiverRefusedLogs" .Values.openTelemetry.prometheus.rules.enabled) }}
+      - alert: JournaldReceiverRefusedLogs
+        expr: sum(rate(otelcol_receiver_refused_log_records_total{receiver=~"journald"}[1m])) > 0
```
timojohlo: Maybe it makes sense to drop the label matcher and have an alert group for all receivers. That means templating {{ $labels.receiver }} into the summary/description so Alertmanager routing and the notification body show which receiver is broken.
 This way we also have alerts for all current and future receivers? This would affect `FilelogRefusedLogs`, `JournaldReceiverRefusedLogs`, `K8sEventsReceiverRefusedLogs` and their playbooks.

timojohlo (changes requested):

timojohlo (approved):

PR #1842, opened 2026-07-23
fix(thanos): rename helper tpl conflict

trouaux:
parent chart over writes the helper function of thanos community helm charts

renaming our tpl to avoid conflict

ref:
https://github.com/thanos-community/helm-charts/blob/master/charts/thanos/templates/_helpers.tpl#L24


ashifnihalb (approved):
LGTM!

PR #1852, opened 2026-07-30
perses: evaluate templates in greenhouse.alertLabels

Frazia04:
## Pull Request Details

Slack alerts emitted by the Perses plugin (`PersesContentSyncJobFailed`, `PersesServiceDown`, `PersesServiceAbsent`) include a `persesDashboard` label whose value is meant to be a direct link to the affected instance's Perses UI. That URL must include the region and TLD (e.g. `https://perses.eu-nl-1.cloud.sap/...`) so the link is clickable from Slack.

Downstream we set this label in the wrapper chart's `values.yaml` using the standard Helm template syntax.

Because the `perses.alertLabels` helper only calls `toYaml` (no `tpl`), the `{{ ... }}` expressions are never evaluated. The rendered `PrometheusRule` ends up with the literal template text in the label value, which produces a broken URL in Slack. The PR fixes it.


ibakshay (approved):
Thank you! LGTM!

PR #1862, opened 2026-08-04
Add transform/falco pipeline

aliyilmaz-deloitte:
## Pull Request Details

This PR adds a transformation pipeline for logs of the falco tool. This PR has been copied from an open [PR](https://github.com/sapcc/helm-charts/pull/12409) in helm-charts. Falco logs existed in the [old OpenSearch instance](https://logs.qa-de-1.cloud.sap/app/data-explorer/discover#?_a=(discover:(columns:!(body),isDirty:!f,sort:!()),metadata:(indexPattern:'audit*',view:discover))&_q=(filters:!(),query:(language:lucene,query:%22falco%22))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))). However, in the new audit.fortlogs or logs.fortlogs instances I couldn't find any of those. 

## Breaking Changes

This doesn't affect the other transformation pipelines. Nevertheless, as no falco logs seem to flow in any of those new instances, we won't be able to validate the functionality of the parsing. The only tests I have conducted were local tests with the same transformation pipeline and original logs from falco. With this setup, everything worked as expected and the fields got parsed correctly.


audit-logs/charts/templates/audit-logs-collector.yaml
```diff
@@ -226,6 +226,55 @@ spec:
...
+            conditions:
+              - resource.attributes["k8s.container.name"] == "falco"
+              - IsMatch(log.body, "^\\{")
+            statements:
+              # proc.aname[N] keys contain brackets which OTTL cannot read from a parsed map;
+              # extract them from the raw JSON string before ParseJSON
+              - merge_maps(log.cache, ExtractGrokPatterns(log.body, "\"proc\\.aname\\[2\\]\"\\s*:\\s*\"(?P<aname2>[^\"]+)\"", true), "upsert")
+              - set(log.attributes["output_fields.proc.aname.2"], log.cache["aname2"]) where log.cache["aname2"] != nil
+              - merge_maps(log.cache, ExtractGrokPatterns(log.body, "\"proc\\.aname\\[3\\]\"\\s*:\\s*\"(?P<aname3>[^\"]+)\"", true), "upsert")
+              - set(log.attributes["output_fields.proc.aname.3"], log.cache["aname3"]) where log.cache["aname3"] != nil
```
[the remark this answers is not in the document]
aliyilmaz-deloitte: The reason this workaround exists (per the comment on line 237) is that proc.aname[N] contains brackets in its JSON key name, which OTTL cannot access after ParseJSON — so these values are extracted from the raw JSON string first, before parsing.
The RE2 regex engine requires exactly (?P<name>...) for named groups — (?<name>...) is not valid syntax.

audit-logs/charts/templates/audit-logs-collector.yaml
```diff
@@ -226,6 +226,55 @@ spec:
...
+              - set(log.attributes["output"], log.cache["output"])
+              - set(log.attributes["priority"], log.cache["priority"])
+              - set(log.attributes["rule"], log.cache["rule"])
+              - set(log.attributes["source"], log.cache["source"])
+              - set(log.attributes["tags"], log.cache["tags"])
+              - set(log.time, Time(log.cache["time"], "%Y-%m-%dT%H:%M:%S.%fZ")) where log.cache["time"] != nil
+              # output_fields
+              - set(log.attributes["output_fields.container.id"], log.cache["output_fields"]["container.id"])
+              - set(log.attributes["output_fields.container.image.repository"], log.cache["output_fields"]["container.image.repository"])
+              - set(log.attributes["output_fields.container.image.tag"], log.cache["output_fields"]["container.image.tag"])
```
olandr, line 256: We probably also want to add container.image.id as a field as well no?

olandr (changes requested):
Nice work, one small comment. And I would guess a version bump would be needed as well.

audit-logs/charts/templates/audit-logs-collector.yaml
aliyilmaz-deloitte: Yes, you're right, I've added it.

olandr (approved):
Looks good!

PR #1867, opened 2026-08-05
feat(kafka): add alerts 

joluc:
## Pull Request Details

Add PrometheusRule alerts for the Kafka plugin, based on monitoring.mixins.dev/kafka.

- 3 alerts: consumer lag too high, under-replicated partitions, no brokers online
- Each alert links to an initial playbook 
- All fire at severity: info for now, with the intended real severity noted inline

### DIFF
--- a/kafka/charts/templates/alerts.yaml
+++ b/kafka/charts/templates/alerts.yaml
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Greenhouse contributors
+# SPDX-License-Identifier: Apache-2.0
+{{- if and (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") .Values.kafka.enabled .Values.monitoring.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ printf "%s-alerts" .Values.kafka.name | trunc 63 }}
+  labels:
+    {{- include "kafka.labels" . | nindent 4 }}
+    {{- with .Values.monitoring.podMonitor.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  groups:
+    - name: kafka.rules
+      rules:
+        - alert: KafkaConsumerLagTooHigh
+          expr: sum by (topic, consumergroup) (kafka_consumergroup_lag{namespace="{{ .Release.Namespace }}", topic!="__consumer_offsets", consumergroup!=""}) > 1000
+          for: 15m
+          labels:
+            severity: info # real severity: warning
+            playbook: https://github.com/cloudoperators/greenhouse-extensions/tree/main/kafka/playbooks/KafkaConsumerLagTooHigh.md
+            {{- include "kafka.alert-labels" . | nindent 12 }}
+          annotations:
+            summary: Kafka consumer lag is too high.
+            description: 'Consumer group {{`{{ $labels.consumergroup }}`}} on topic {{`{{ $labels.topic }}`}} has a lag of {{`{{ printf "%.0f" $value }}`}} messages.'
+        - alert: KafkaUnderReplicatedPartitions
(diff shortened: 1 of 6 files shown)


kafka/charts/templates/alerts.yaml
```diff
@@ -0,0 +1,47 @@
...
+    {{- include "kafka.labels" . | nindent 4 }}
+    {{- with .Values.monitoring.podMonitor.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  groups:
+    - name: kafka.rules
+      rules:
+        - alert: KafkaConsumerLagTooHigh
+          expr: sum by (topic, consumergroup) (kafka_consumergroup_lag{topic!="__consumer_offsets", consumergroup!=""}) > 1000
```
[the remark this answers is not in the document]
joluc: Added

kafka/playbooks/KafkaConsumerLagTooHigh.md
```diff
@@ -0,0 +1,36 @@
+---
+title: KafkaConsumerLagTooHigh
+weight: 20
+---
+
+# KafkaConsumerLagTooHigh
+
+## Problem
+
+A consumer group's lag on a topic has exceeded 1000 messages for more than 15 minutes. Consumers are falling behind the producers.
```
olandr, line 10: I think this is a bit of an confusing sentence. What does it mean that Cs are _falling behind_ Ps?

kafka/playbooks/KafkaNoBrokersOnline.md
```diff
@@ -0,0 +1,45 @@
+---
+title: KafkaNoBrokersOnline
+weight: 20
+---
+
+# KafkaNoBrokersOnline
+
+## Problem
+
+The `kafka_server_kafkaserver_brokerstate` metric is entirely absent, meaning no Kafka broker is reporting metrics. The cluster is almost certainly down or unreachable from Prometheus.
```
olandr, line 10: How is this different from absent-metrics alerts? I remember we having those a lot before I left, is it the same or am I forgetting something here?

kafka/playbooks/KafkaNoBrokersOnline.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Diagnosis
+
+1. Check the broker pods:
+
+   ```
+   kubectl get pods -n <kafka-namespace> -l strimzi.io/component-type=kafka -o wide
+   ```
+
+   - No pods → the `Kafka` resource or namespace was removed.
```
olandr: _ was removed._ to me sounds like it was someone who actively removed it. I would phrase it more as: no pods -> kafka is not deployed or failing to schedule.

To me that sounds like a more reasonable reason for why they are not present, and hints an operator to figure out why they could be gone. Ofc it could still be that the resource or namespace was removed (by a human or something else), but it would be quite odd imo.

olandr (changes requested):

kafka/playbooks/KafkaConsumerLagTooHigh.md
joluc: Kafka terminology. We produce more than we consume

kafka/playbooks/KafkaNoBrokersOnline.md
joluc: I was thinking about it but here we can directly set it to critical to have a onduty call. Absent is just a warning if i remember well? Or maybe I need to read more about that absent metric

kafka/playbooks/KafkaNoBrokersOnline.md
joluc: Reworded

kafka/playbooks/KafkaNoBrokersOnline.md
olandr: Hmm, I guess that makes sense to have that control here. but yeah I do not remember what the absent ones are  doing.

olandr (approved):

PR #1875, opened 2026-08-06
feat(audit-logs): translate old audit logstash into otelcol

olandr:
This should add the conversion of old logstash log shippers into the new modern otelcol format.

Adds three new external inputs:
- syslog
- http
- mtls

---------


olandr:
pivot.

PR #1880, opened 2026-08-07
feat(logs): add HTTP-JSON receiver on external collector

joluc:
## Pull Request Details

add HTTP-JSON receiver on external collector


logs/charts/templates/_external-http-config.tpl
```diff
@@ -0,0 +1,51 @@
...
+webhookevent/external-http:
+  endpoint: "0.0.0.0:{{ .Values.openTelemetry.externalCollector.externalHttpConfig.port }}"
+  path: {{ .Values.openTelemetry.externalCollector.externalHttpConfig.path | quote }}
+  health_path: {{ printf "%s/health" .Values.openTelemetry.externalCollector.externalHttpConfig.path | quote }}
+  split_logs_at_newline: false
+{{- if .Values.openTelemetry.externalCollector.externalHttpConfig.tls.enabled }}
+  tls:
+    cert_file: /etc/ssl/syslog-tls/tls.crt
+    key_file: /etc/ssl/syslog-tls/tls.key
+{{- end }}
```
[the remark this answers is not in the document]
joluc: Yes, valid concern. The receiver is intentionally open in this iteration.

logs/charts/ci/test-values-all-features.yaml
```diff
@@ -95,6 +95,13 @@ openTelemetry:
       enabled: true
       otlp_grpc_port: 4317
       otlp_http_port: 4318
+    externalHttpConfig:
+      enabled: true
+      port: 1517
+      path: /audit/external
+      forwardedBy: external-http
+      tls:
+        enabled: true
```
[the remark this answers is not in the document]
joluc: Fair point that the all-features fixture doesn't isolate the HTTP-only non-Kafka path. Keeping CI to the existing two fixtures for now

logs/charts/templates/_external-http-config.tpl
```diff
@@ -0,0 +1,51 @@
...
+
+{{- define "external_http.transform" }}
+transform/external-http:
+  error_mode: ignore
+  log_statements:
+    - context: log
+      statements:
+        - set(log.time_unix_nano, log.observed_time_unix_nano)
+        - merge_maps(log.attributes, ParseJSON(log.body), "upsert") where IsMatch(log.body, "^\\{")
```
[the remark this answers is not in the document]
joluc: Correct, this is intentional for the current sender approach. The octobus Logstash http output uses format => "json" (one JSON object per request), verified end-to-end. Array/json_lines bodies are accepted but stored as a single record with the raw body preserved, no data loss, just no per-record splitting. Documented this in the path value comment

logs/charts/templates/_external-audit-exporter.tpl
```diff
@@ -0,0 +1,146 @@
+{{/*
+SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Greenhouse contributors
+SPDX-License-Identifier: Apache-2.0
+*/}}
+{{/*
+  ============================================================================
+  Shared audit exporters, connectors, extensions and pipelines that write to
+  the audit stream (OpenSearch audit-datastream, or the Kafka auditKafkaTopic).
+
+  Two sources feed the same audit destination: the syslog audit path
```
olandr, line 10: Why is basically everything in this file named syslog? Is this not solely for the http json external receiver?

logs/charts/ci/test-values-all-features.yaml
```diff
@@ -95,6 +95,13 @@ openTelemetry:
       enabled: true
       otlp_grpc_port: 4317
       otlp_http_port: 4318
+    externalHttpConfig:
+      enabled: true
+      port: 1517
```
olandr, line 100: Small nit pick. What is the choice for this port? Just curious why use a registered one? But I would have considered an alt http or https.

Also are we able to send to specific ports in the logstash forwarding?

logs/charts/ci/test-values-all-features.yaml
olandr, line 104: I do not understand the copilot here.

Why are we setting tls? How are we actually terminating it? Somewhere I saw that we will run syslogTLS certs for terminating. Not a problem per se I think, but the naming is really off. SyslogTLS certs for an https webhook? 


I think we really need to solve this naming directly after launch, it is just way too confusing and also prone to errors down the line.

logs/charts/templates/_helpers.tpl
```diff
@@ -5,6 +5,30 @@ Generic plugin name
...
+
+{{/*
+True when the collector must terminate TLS with the logs-syslog-tls cert —
+either syslog-TLS ingestion or the HTTP-JSON receiver's TLS is on. Both reuse
+the same cert-manager Certificate.
+Usage: {{- if eq (include "externalCollector.tlsEnabled" .) "true" }}
+*/}}
+{{- define "externalCollector.tlsEnabled" -}}
+{{- if or .Values.openTelemetry.externalCollector.syslogTLSConfig.enabled (and .Values.openTelemetry.externalCollector.externalHttpConfig.enabled .Values.openTelemetry.externalCollector.externalHttpConfig.tls.enabled) -}}
+true
```
olandr, line 28: I guess this is the same as my previous question. Quite confusing that both syslogtls and externalhttpconfig.tls needs to be activated for tls.

I would want to rename this syslogtls. But in the interim I am fine with a fixme and we deal with it later. (Although postponing renaming we would need to qa and roll out new secrets then for the prod cluster after golive...)

logs/charts/ci/test-values-all-features.yaml
olandr: I just checked, this is simply an increment from the other alert, deploy ports 1515 1516

logs/charts/templates/_external-audit-exporter.tpl
olandr: I will let this slide for now, I saw that blocks of code were moved from the old syslog-centred setup. Changing names now would scope creep somewhat (although possibly being a more sensible naming down the line).

logs/charts/templates/_helpers.tpl
olandr: Config wise in the interm we just need all of these true if we want to have tls I suppose

olandr (approved):
Had another review, things needs a new name but for the time being I am fine with this.

PR #1886, opened 2026-08-10
fix(logs): adds extra conditions for safe parsing of syslog and non syslog messages

olandr:
due to the nature of some messages being empty, a cond has been added. also, for some messages that are not parsable by syslog are now routed safely and tagged with an unkown format

---------


joluc (approved):

joluc:
version bump missing

PR #1889, opened 2026-08-11
Classify network logs as audit-relevant

aliyilmaz-deloitte:
## Pull Request Details
- Adds a new audit classification rule in transform/syslog_audit_classification that marks syslog logs from network security devices as audit-relevant by matching against the log body.
- Covered vendors: ISE-IDC, Check Point, Fortinet, Palo Alto Networks, TrendMicro, Tufin.
## Breaking Changes
X

### DIFF
--- a/digicert-issuer/plugindefinition.yaml
+++ b/digicert-issuer/plugindefinition.yaml
@@ -6,14 +6,14 @@ kind: PluginDefinition
 metadata:
   name: digicert-issuer
 spec:
-  version: 1.3.3
+  version: 1.4.0
   displayName: DigiCert extensions
   description: Extensions to the cert-manager for DigiCert support
   docMarkDownUrl: https://raw.githubusercontent.com/cloudoperators/greenhouse-extensions/main/digicert-issuer/README.md
   helmChart:
     name: digicert-issuer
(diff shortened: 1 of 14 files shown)


melvinsimmert (approved):

olandr (approved):

melvinsimmert:
This will forward the data to audit opensearch, but we will still miss the "sap.cc.audit.source"-field in the logs. I think this is a result we can take for now, but we should add the relevant field for all of these logs ASAP afterwards.

@aliyilmaz-deloitte Can you bump the versions again?

melvinsimmert (changes requested):
Please version bump to 0.4.39

olandr (dismissed):

melvinsimmert (approved):

olandr (dismissed):

olandr:
~Sorry for being so slow with the approval. I will bump the chart this time.~


edit: ah you forked it ofc, then you need to bump it again to 0.4.40

melvinsimmert (changes requested):
As Simon pointed out, another version bump is necessary.

olandr (approved):

PR #1910, opened 2026-08-19
fix(logs): adds condition on empty message to stop warnings

olandr:
---------


olandr:
stale.

PR #1936, opened 2026-08-26
fix(logs,audit-logs): moves ingester field normalisation rules to external

olandr:
---------


olandr:
Needs #1935

timojohlo (approved):

PR #1980, opened 2026-09-07
feat(logs): enable mount of kafka CA cert

ztomaszewska:
Enable to mount kafka brokers CA cert in otel collector so that we can verify TLS connection for internal traffic.

### DIFF
--- a/logs/charts/templates/ingester-collector.yaml
+++ b/logs/charts/templates/ingester-collector.yaml
@@ -54,7 +54,7 @@ spec:
 {{- if and $cfg.kafka (gt (len (default (list) $cfg.kafka.brokers)) 0) }}
 {{- $kafka = $cfg.kafka }}
 {{- end }}
-{{- $kafkaCASecret := dig "tls" "caSecret" "" $kafka }}
+{{- $kafkaCASecret := $kafka.tls.caSecret }}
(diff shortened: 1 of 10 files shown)


logs/charts/templates/ingester-collector.yaml
```diff
@@ -73,6 +85,9 @@ spec:
 {{- if or (default false (dig "tls" "enabled" false $kafka)) $root.Values.openTelemetry.kafka.tls.enabled }}
         tls:
           insecure: false
+{{- if $shouldMountKafkaCASecret }}
+          ca_file: /etc/ssl/kafka/{{ dig "tls" "caSecretKey" "" $kafka }}
```
olandr: A bit inconsistent with the assigning of ca_file in this PR.

Here you do `ca_file: /etc/ssl/kafka/{{ dig "tls" "caSecretKey" "" $kafka }}` but above you do `ca_file: /etc/ssl/kafka/{{ .Values.openTelemetry.kafka.tls.caSecretKey }}`. Does it make sense to keep it consistent? I would opt for `ca_file: /etc/ssl/kafka/{{ .Values.openTelemetry.kafka.tls.caSecretKey }}` as it reduces number of variables.

olandr (changes requested):

logs/charts/templates/ingester-collector.yaml
ztomaszewska: I changed it to `$kafka.tls.caSecretKey` as we operate here on `$kafka` and not `Values.openTelemetry.kafka`

olandr (approved):

PR #1999, opened 2026-09-14
fix(logs): external otelcol should use extrnal kafka port so that TLS connection can be verified (kafka identity).

ztomaszewska:
Signed-off-by: Zuzanna Tomaszewska <zuzanna.tomaszewska@sap.com>


logs/charts/Chart.yaml
```diff
@@ -4,7 +4,7 @@
 apiVersion: v2
 appVersion: 0.152.0
 name: logs
-version: 0.4.65
+version: 0.4.66
```
olandr: Need to bump this to 67.

olandr (changes requested):

logs/charts/Chart.yaml
ztomaszewska: done

olandr (approved):

PR #2018, opened 2026-09-17
fix(gatekeeper): exempt system namespaces from the admission webhooks

mikolajkucinski:
## Pull Request Details

Exempt kube-system and the other system namespaces from the gatekeeper validation and mutation webhooks by default, so gatekeeper stops evaluating them (Gardener best-practice for webhooks that could otherwise block kube-system). gatekeeper-system stays in the exempt list, otherwise the chart merge drops the mandatory self-exemption on `kubernetes.io/metadata.name`.

Does not cover `check-ignore-label.gatekeeper.sh`, its selector is hardcoded in the upstream OPA chart so Gardener keeps remediating that one.

## Breaking Changes

None.

## Issues Fixed

None.

## Other Relevant Information

Chart default, so it applies to every cluster using this plugin, not per-cluster.


mikolajkucinski:
closing this, we'll scope the exempt to the obs preset instead of a chart-wide default (per the thread with abhi)

PR #2034, opened 2026-09-21
fix(.github): changes E2E flux helm to check against PR target branch instead of main.

olandr:
There is an issue with the e2e flux testing ci when you use stacked PRs. This would change the behaviour to compare no of plugins changed with the target branch (stack base) instead of main.

---------


abhijith-darshan (approved):

olandr:
For some reason not working as expected. needs more testing.

